PR #7, opened 2015-06-18
Push to pulp

midnightercz:
new push_to_pulp functions and fixes related to them.

### DIFF
--- a/bin/dock-pulp.py
+++ b/bin/dock-pulp.py
@@ -385,6 +385,10 @@ def do_push_to_pulp(bopts, bargs):
     #    help='format the json into something human-readable')
     #opts, args = parser.parse_args(bargs)
     parser = OptionParser(usage=do_push_to_pulp.__doc__)
+    parser.add_option('-d', '--desc',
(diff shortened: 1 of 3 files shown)


TomasTomecek:
can this be merged please? this is what dock uses to push to pulp https://github.com/DBuildService/dock/blob/626314e8dea3078d2e034a8639245431280aa69b/dock/plugins/post_push_to_pulp.py#L114

Zyzyx:
No for 3 reasons:
* This should have been implemented as extensions to upload(). This introduces an entirely separate code path.
* I do not understand the difference between get_id() and get_top_layer() in imgutils.  	6a35ab5 introduces it, and it sounds like it is meant to get the origin layer (the oldest parent). In 61d4994 it is changed to get the "youngest" layer, which is what get_id already does.
* I'm against arbitrarily setting the docker ID in create(). That makes enforcing what the docker IDs should be relative to the Pulp IDs impossible, or at least defers it to another system.

Zyzyx:
I'll propose a patch set that is more maintainable this afternoon.

PR #9, opened 2015-07-01
crane() improved preformace

midnightercz:
added wait=True as default crane() parameter.
With wait=false, crane method won't wait for tasks to finish and returns list of spawned tasks instead. User can then call dockpulp.watch_tasks(tasks) which will wait print status of finished tasks and wait until all tasks finish.
With wait=True, method will behave as in original dockpulp

This PR affects only performance, it's not critical and doesn't include bug fixes.


Zyzyx:
I still reserve the right to revisit that push_tar_to_pulp command later! I confess I won't find the time right now, and the rebase is appreciated it.

TomasTomecek:
Thanks for merging this!

PR #12, opened 2015-07-03
Add simplejson as a requirement

twaugh:
The project requires `simplejson`, which may not be installed. Here is one approach for handling that: deal with it in `setup.py`.

The other approach would be to use `json` instead of `simplejson`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,6 +2,13 @@
 import sys
 from setuptools import setup, Command
 
+def _simplejson_on_python26():
+    if (sys.version_info[0] > 2 or
+        (sys.version_info[0] == 2 and
+         sys.version_info[1] > 6)):
+        return []
+    return ['simplejson']
+
 setup(
     name = "dockpulp",
     version = "1.11",
@@ -12,6 +19,7 @@ setup(
     url = "https://github.com/release-engineering/dockpulp.git",
     package_dir = {'': '.'},
     packages = ['dockpulp'],
+    install_requires = _simplejson_on_python26(),
     scripts = ['bin/dock-pulp.py', 'bin/dock-pulp-bootstrap.py', 'bin/dock-pulp-restore.py'],
     package_data = {'': ['conf/dockpulp.conf']},
 )


Zyzyx:
Can you elaborate on why you want to open a file instead of put the contents directly in setup.py?

simplejson is required for running on RHEL 6. json works fine for RHEL 7 and later.

twaugh:
It was to be extensible. It can go directly in setup.py if you'd rather.

Zyzyx:
How would you propose handling conditional requirements in setup.py? It'd be nice if things built with the same code on either RHEL 6 or 7. In the spec file I get away with it like so:
```
%if 0%{?rhel} == 6
Requires: python-simplejson
%endif
```

I'm trying to avoid branching.

twaugh:
simplejson is required on RHEL 6 because the built-in json module has bugs? Because `dockpulp/__init__.py` currently imports json, and only tries to import simplejson if importing json failed.

```
try:
    import json
except ImportError:
    # Python 2.6 and earlier
    import simplejson as json
```

```
Python 2.6.6 (r266:84292, Nov 21 2013, 10:50:32) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-4)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import json
>>> open("/etc/redhat-release").read()
'Red Hat Enterprise Linux Workstation release 6.6 (Santiago)\n'
>>> import sys; sys.version
'2.6.6 (r266:84292, Nov 21 2013, 10:50:32) \n[GCC 4.4.7 20120313 (Red Hat 4.4.7-4)]'
>>> 
```

We could have `setup.py` check `sys.version_info` and only require simplejson on Python < 2.7. We'd have to also fix up files that import simplejson without catching `ImportError` though.

Zyzyx:
Rats. I have to switch that around. The json built-in in RHEL 6 does not have the same behaviors as simplejson. In Python 2.7 (RHEL 7) the built-in json library matched simplejson.

twaugh:
See https://github.com/twaugh/dockpulp/commit/1cc4cecfd487e39e77cd300d2bf63af95885b83e.

Not sure if it's normal practice.

PR #16, opened 2015-07-20
Require simplejson on Python 2.6

twaugh:
For Python > 2.6, just prefer it.

This is based on https://github.com/release-engineering/dockpulp/pull/12#issuecomment-119216836 "Rats. I have to switch that around."


TomasTomecek:
+1

to avoid copy-paste, the code could live on one place and the import could be done as
```python
from dockpulp import json
```
not sure if that's best practice or not

PR #24, opened 2015-11-03
Port certificate-based auth from Jindra's fork

mmilata:
@midnightercz, are you OK with bringing this into master?

### DIFF
--- a/bin/dock-pulp.py
+++ b/bin/dock-pulp.py
@@ -88,16 +88,21 @@ def pulp_login(bopts):
     p = dockpulp.Pulp(env=bopts.server, config_file=bopts.config_file)
     if bopts.debug:
         p.setDebug()
-    if bopts.cert and bopts.key:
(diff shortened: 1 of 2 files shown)


bin/dock-pulp.py
```diff
@@ -88,16 +88,21 @@ def pulp_login(bopts):
     p = dockpulp.Pulp(env=bopts.server, config_file=bopts.config_file)
     if bopts.debug:
         p.setDebug()
-    if bopts.cert and bopts.key:
+    if bopts.cert:
         p.certificate = bopts.cert
+    if bopts.key:
         p.key = bopts.key
```
twaugh: Should this be using `p.set_certs()` instead now?

midnightercz:
I guess it should be ok to cherry-pick this one

breillyrh:
Going to pull this in then.

PR #25, opened 2015-11-03
watch_tasks implementation that returns when any task fails

mmilata:
I've added a commit that fixes some incorrect information in comments and removes the old function.

@midnightercz, OK to port this to master?


midnightercz:
looks good

PR #28, opened 2015-11-09
syncRepo: enforce prefix

twaugh:
The caller doesn't know what the prefix needs to be, so `syncRepo()` should add it if necessary (like `createRepo()` does).

### DIFF
--- a/dockpulp/__init__.py
+++ b/dockpulp/__init__.py
@@ -729,8 +729,12 @@ class Pulp(object):
         """turn on debug output"""
         log.setLevel(logging.DEBUG)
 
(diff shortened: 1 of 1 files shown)


TomasTomecek:
LGTM (sync repo and create repo should have similar logic)

breillyrh:
I wasn't aware createRepo in the master branch had the same logic here. I'm going to merge this change for now, but I think the prefix should be configurable in /etc/dockpulp.conf - not hardcoded as 'redhat-'.

twaugh:
Totally agree, it should be configurable.

TomasTomecek:
+1

PR #43, opened 2016-02-29
Pulp.upload(): new chunk_size parameter, defaults to 5Mb

twaugh:


twaugh:
Note this default requires a modification to SSLRenegBufferSize on the server.

breillyrh:
Implemented this change using a slightly different method, NACKing.

PR #48, opened 2016-03-10
syncRepo: don't change the sync environment (#47)

### LINKED ISSUE #47: syncRepo cannot sync from https:// registry feed
This code is wrong:

```
        if syncenv.startswith("https"):
            syncenv = syncenv.replace("https","http")
```

https://github.com/release-engineering/dockpulp/blob/master/dockpulp/__init__.py#L754

Whatever is in the config should be used as-is, without changing port numbers or schemes.

twaugh:
Fixes #47.

### DIFF
--- a/dockpulp/__init__.py
+++ b/dockpulp/__init__.py
@@ -751,9 +751,6 @@ class Pulp(object):
 
         syncenv = self.syncenv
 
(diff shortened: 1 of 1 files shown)


twaugh:
Thanks!

PR #49, opened 2016-04-13
Check if layers already exists on the server

vrutkovs:
This would help atomic-reactor upload less layers


breillyrh:
Looks good to me.

PR #50, opened 2016-04-28
Added in support for distributor interactions

breillyrh:


Zyzyx:
Ugh, passing in json structures on the command line, really? This is just begging for user error when it is missing a backslash or something.

breillyrh:
Fair point, but distributor configs can vary wildly. Do you have a suggestion for an alternative solution?

breillyrh:
Changed it so that associate depends on dockpulpdistributors.json.

PR #51, opened 2016-05-19
Dock-Pulp confirm for protected repos

breillyrh:
Dock-pulp can now confirm protected repositories, with two caveats.

I had to use subprocess / curl instead of the more preferable python requests because requests does not support encrypted keys. 
Also, the way our entitlements work require the user to be root. What this means for dockpulp is that you'll have to log into the environment as root and run confirm as root for protected repos.

### DIFF
--- a/bin/dock-pulp.py
+++ b/bin/dock-pulp.py
@@ -22,6 +22,7 @@ import shutil
 import sys
 import logging
 import ast
+import subprocess
 
 try:
     # Python 2.6 and earlier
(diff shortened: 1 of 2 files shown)


Zyzyx:
Why is root required?

breillyrh:
It's not something inherent to dockpulp, I was just unable to use entitlement certs as anything other than root. It's possible I was making a mistake on my end during testing, but if I was it would not require a code change here. "curl: (58) unable to load client key: -8178 (SEC_ERROR_BAD_KEY)"

breillyrh:
It turns out it was user error, root is not needed. I just had to change the permissions on my .cert and .key files.

PR #52, opened 2016-05-26
Updated confirm to use requests and silent option

breillyrh:
Removed need for curl in confirm function, now accepts cert and key and uses python requests. Added in machine readable output for confirm as a --silent option.


Zyzyx:
I would recommend define a "log" function of your own that has the "if not silent" conditional in it so you don't have them sprinkled everywhere taking up space. Good work though!

breillyrh:
Good point, I've removed the vast majority of bool checks.

PR #55, opened 2016-06-21
Updated sync to allow auth

breillyrh:
This is to fulfill atomic reactor requirements.

### DIFF
--- a/dockpulp/__init__.py
+++ b/dockpulp/__init__.py
@@ -829,28 +829,46 @@ class Pulp(object):
         """turn on debug output"""
         log.setLevel(logging.DEBUG)
 
-    def syncRepo(self, env, repo, config_file=DEFAULT_CONFIG_FILE,
-                 prefix_with="redhat-"):
+    def syncRepo(self, env=None, repo=None, config_file=DEFAULT_CONFIG_FILE,
+                 prefix_with="redhat-", feed=None, upstream_name=None,
+                 basic_auth_username=None, basic_auth_password=None,
+                 ssl_validation=None):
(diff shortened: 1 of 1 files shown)


twaugh:
LGTM, not yet tested against a registry that requires auth.

dockpulp/__init__.py
```diff
@@ -829,28 +829,46 @@ def setDebug(self):
...
-
-        syncenv = self.syncenv
-
-        data = {
-            'override_config': {
-                'ssl_validation': False,
-                'feed': syncenv,
-                'upstream_name': repoid
+        if not upstream_name:
+            repoinfo = self.listRepos(repo, True)
```
twaugh: Oh, the method tries to return repoinfo so if upstream_name is not None, that fails.

Maybe syncRepo could return the tid instead?

twaugh: Hmm, that's no good, because the caller needs to know the actual repo name used, including any prefix.

twaugh: Works fine as long as I don't provide upstream_name, so I think we should drop that parameter and just always call listRepos().

breillyrh: Yes, one of the reasons I list out repo information there is to see what images the repo has currently vs what it will have afterwards, so it can copy the new images to redhat-everything. That's done in the cli code though, so I'll have to bring that over here so that atomic-reactor will also copy images to redhat-everything.

breillyrh: I'm going to go ahead and merge this now and remove the upstream_name parameter in the next patch.

Issue #56, opened 2016-06-21
Provide a way to discover the default prefix


twaugh:
dockpulp adds a default prefix to repo IDs, but callers have no way to find out what it is.

This is a problem when it comes to e.g. creating repositories on demand, because the straightforward route doesn't work:

1. Ask for list of existing repo IDs
2. Look through the list for the one we need
3. If not in the list, create it

This is because the list of existing repo IDs (`getRepos([...], fields=["id"])`) includes the prefix, but the repo id we know we want to create does not.

One solution would be for dockpulp to have a global PREFIX variable, so callers could inspect that. Now the solution to creating on demand becomes:

1. Prefix the repo ID we have with dockpulp.PREFIX
2. Look for that repo ID
3. If not found, create it

breillyrh:
Added in https://github.com/release-engineering/dockpulp/commit/60e27193615428a663b42d258ca7b26afa15e388

PR #59, opened 2016-06-22
Updated sync to always copy new images and manifests to HIDDEN repo

breillyrh:
syncRepo now returns a tuple of new images and manifests instead of old repo information.

### DIFF
--- a/dockpulp/__init__.py
+++ b/dockpulp/__init__.py
@@ -893,7 +893,8 @@ class Pulp(object):
                     'feed': feed,
                     'upstream_name': upstream_name,
                     'basic_auth_username': basic_auth_username,
-                    'basic_auth_password': basic_auth_password
+                    'basic_auth_password': basic_auth_password,
+                    'enabled_v1': True
                 }
             }
         else:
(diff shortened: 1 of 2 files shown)


dockpulp/__init__.py
```diff
@@ -911,7 +913,36 @@ def syncRepo(self, env=None, repo=None, config_file=DEFAULT_CONFIG_FILE,
...
+            oldimgs = []
+        else:
+            oldimgs = repoinfo['images'].keys()
+            
+        if len(repoinfo['manifests'].keys()) == 0:
+            oldmanifests = []
+        else:
+            oldmanifests = repoinfo['manifests'].keys()
+
+        repoinfo = self.listRepos(repo, True)
```
twaugh: Oh, you need to do it twice to see which units are new, I see now.

twaugh:
Now that syncRepo() no longer returns the repository ID I've adjusted atomic-reactor's pulp_sync plugin so that it doesn't rely on that. Instead it finds out the prefix from a configuration parameter and works out the repository ID from that. Even better would be for dockpulp to be able to tell it the prefix (https://github.com/release-engineering/dockpulp/issues/56).

twaugh:
LGTM

PR #60, opened 2016-06-27
Updated confirm for v2 manifests and blobs

breillyrh:
New option to specifically test v1 or v2 content.

### DIFF
--- a/bin/dock-pulp.py
+++ b/bin/dock-pulp.py
@@ -428,18 +428,18 @@ def _test_repo(dpo, dockerid, redirect, pulp_imgs, protected=False, cert=None, k
     result['missing_ancestor_layers'] = missing
     return result
 
-def _test_repoV2(dpo, dockerid, redirect, pulp_manifests, pulp_blobs, protected=False, cert=None, key=None, silent=False):
+def _test_repoV2(dpo, dockerid, redirect, pulp_manifests, pulp_blobs, pulp_tags, protected=False, cert=None, key=None, silent=False):
     """confirm we can reach crane and get data back from it"""
     result = {}
     result['error'] = False
     url = dpo.registry + '/v2/' + dockerid + '/manifests'
     log.info('  Testing Pulp and Crane manifests')
-    log.debug('  contacting %s' % url)
-    c_manifests = []
+    log.debug('  contacting %s', url)
+    c_manifests = set([])
     if protected:
         log.info('  Repo is protected, trying certs')
         answer = requests.get(url, verify=False)
-        if answer.content != 'Not Found':
+        if answer.status_code != requests.codes.not_found:
             log.warning('  Crane not reporting 404 - possibly unprotected?')
         if cert is None and key is None:
             log.error('  Must provide a cert to test protected repos, skipping')
@@ -449,10 +449,10 @@ def _test_repoV2(dpo, dockerid, redirect, pulp_manifests, pulp_blobs, protected=
     try:
         for manifest in pulp_manifests:
             answer = requests.get(url + '/' + manifest, verify=False, cert=(cert,key))
-            log.debug('  crane content: %s' % answer.content)
-            log.debug('  status code: %s' % answer.status_code)
-            if answer.status_code == 200:
-                c_manifests.append(manifest)
+            log.debug('  crane content: %s', answer.content)
+            log.debug('  status code: %s', answer.status_code)
+            if answer.status_code == requests.codes.ok:
+                c_manifests.add(manifest)
 
     except requests.exceptions.SSLError:
         log.error('  Request failed due to invalid cert or key')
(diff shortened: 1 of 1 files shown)


bin/dock-pulp.py
```diff
@@ -428,6 +428,133 @@ def _test_repo(dpo, dockerid, redirect, pulp_imgs, protected=False, cert=None, k
...
+    except requests.exceptions.SSLError:
+        log.error('  Request failed due to invalid cert or key')
+        result['error'] = True
+        return result
+        
+    p_manifests = set(pulp_manifests)
+    c_manifests = set(c_manifests)
+
+    pdiff = p_manifests - c_manifests
+    cdiff = c_manifests - p_manifests
```
twaugh: I guess c_manifests can't ever have items not in p_manifests here. Actually I'm not even sure how to get a list of manifest digests from Crane... you can get a list of tags for the repository, but I don't think you can resolve a tag into a manifest digest because Crane doesn't set the Docker-Content-Digest header correctly when fetching a manifest by tag.

bin/dock-pulp.py
```diff
@@ -428,6 +428,133 @@ def _test_repo(dpo, dockerid, redirect, pulp_imgs, protected=False, cert=None, k
...
+    log.debug('  crane manifests: %s' % c_manifests)
+    log.debug('  pulp manifests: %s' % p_manifests)
+    same = True
+    for p_manifest in p_manifests:
+        if p_manifest not in c_manifests:
+            same = False
+    for c_manifest in c_manifests:
+        if c_manifest not in p_manifests:
+            same = False
+    if not same:
```
twaugh: Instead of calculating `same`, wouldn't `if pdiff or cdiff:` work here?

bin/dock-pulp.py
```diff
@@ -428,6 +428,133 @@ def _test_repo(dpo, dockerid, redirect, pulp_imgs, protected=False, cert=None, k
...
+    except requests.exceptions.SSLError:
+        log.error('  Request failed due to invalid cert or key')
+        result['error'] = True
+        return result
+        
+    p_blobs = set(pulp_blobs)
+    c_blobs = set(c_blobs)
+
+    pdiff = p_blobs - c_blobs
+    cdiff = c_blobs - p_blobs
```
twaugh: Same comment as above: I don't think cdiff will ever have anything in.

bin/dock-pulp.py
```diff
@@ -428,6 +428,133 @@ def _test_repo(dpo, dockerid, redirect, pulp_imgs, protected=False, cert=None, k
...
+    log.debug('  crane blobs: %s' % c_blobs)
+    log.debug('  pulp blobs: %s' % p_blobs)
+    same = True
+    for p_blob in p_blobs:
+        if p_blob not in c_blobs:
+            same = False
+    for c_blob in c_blobs:
+        if c_blob not in p_blobs:
+            same = False
+    if not same:
```
twaugh: Same comment as above: I think we've already calculated `same` in `pdiff` and `cdiff`.

bin/dock-pulp.py
```diff
@@ -428,6 +428,133 @@ def _test_repo(dpo, dockerid, redirect, pulp_imgs, protected=False, cert=None, k
...
+            same = False
+    for c_manifest in c_manifests:
+        if c_manifest not in p_manifests:
+            same = False
+    if not same:
+        pdiff = ', '.join((p_manifests - c_manifests))
+        cdiff = ', '.join((c_manifests - p_manifests))
+        
+        log.error('  Pulp manifests and Crane manifests are not the same:')
+        if len(pdiff) > 0:
```
twaugh: It's generally better to write `if pdiff:`, which checks whether there are any items in pdiff at all, rather than calculating exactly how many there are and comparing with zero. I see `len(...) > 0` elsewhere in this function too.

bin/dock-pulp.py
```diff
@@ -428,6 +428,133 @@ def _test_repo(dpo, dockerid, redirect, pulp_imgs, protected=False, cert=None, k
     result['missing_ancestor_layers'] = missing
     return result
 
+def _test_repoV2(dpo, dockerid, redirect, pulp_manifests, pulp_blobs, protected=False, cert=None, key=None, silent=False):
+    """confirm we can reach crane and get data back from it"""
+    result = {}
+    result['error'] = False
+    url = dpo.registry + '/v2/' + dockerid + '/manifests'
+    log.info('  Testing Pulp and Crane manifests')
+    log.debug('  contacting %s' % url)
```
twaugh: Logging functions can take arguments for %-substitution, to avoid unnecessarily formatting strings if the log level would omit this message, so this could be written: 
```
log.debug('  contacting %s', url)
```

bin/dock-pulp.py
```diff
@@ -428,6 +428,133 @@ def _test_repo(dpo, dockerid, redirect, pulp_imgs, protected=False, cert=None, k
...
+    c_manifests = []
+    if protected:
+        log.info('  Repo is protected, trying certs')
+        answer = requests.get(url, verify=False)
+        if answer.content != 'Not Found':
+            log.warning('  Crane not reporting 404 - possibly unprotected?')
```
twaugh: Should this be `if answer.status_code != requests.codes.not_found:`?

bin/dock-pulp.py
```diff
@@ -428,6 +428,133 @@ def _test_repo(dpo, dockerid, redirect, pulp_imgs, protected=False, cert=None, k
...
+        if cert is None and key is None:
```
twaugh: Should this be `or`?

bin/dock-pulp.py
```diff
@@ -428,6 +428,133 @@ def _test_repo(dpo, dockerid, redirect, pulp_imgs, protected=False, cert=None, k
...
     return result
 
+def _test_repoV2(dpo, dockerid, redirect, pulp_manifests, pulp_blobs, protected=False, cert=None, key=None, silent=False):
+    """confirm we can reach crane and get data back from it"""
+    result = {}
+    result['error'] = False
+    url = dpo.registry + '/v2/' + dockerid + '/manifests'
+    log.info('  Testing Pulp and Crane manifests')
+    log.debug('  contacting %s' % url)
+    c_manifests = []
```
twaugh: We only use c_manifests as a set, so we could declare it as one here and use `.add` instead of `.append`.

bin/dock-pulp.py
breillyrh: Would there ever be a case where it would make sense to pass a key and not a cert, or a cert without a key?

bin/dock-pulp.py
breillyrh: Yes this is true... and for the purposes of confirm it shouldn't matter, I think. Confirm only checks if all the content in a repo is reachable, though it would be nice to see exactly what Crane has.

bin/dock-pulp.py
twaugh: I guess I don't know enough about it. Maybe `and` is right.

bin/dock-pulp.py
twaugh: ack

twaugh:
LGTM

PR #61, opened 2016-06-28
Confirm uses default pulp redirect correctly

breillyrh:


bin/dock-pulp.py
```diff
@@ -1028,14 +949,17 @@ def do_sync(bopts, bargs):
     dock-pulp sync [options] <env to sync from> repo-id
     Sync a repo from one environment to another"""
     parser = OptionParser(usage=do_sync.__doc__)
+    parser.add_option('-p', '--password', default=None, help='specify a password')
```
lcarva: FYI `default=None` is redundant, as that's its default value: https://docs.python.org/3/library/argparse.html#default

bin/dock-pulp.py
```diff
@@ -214,139 +215,59 @@ def _test_repo(dpo, dockerid, redirect, pulp_imgs, protected=False, cert=None, k
...
-        reachable = list(reachable)
-        reachable.sort()
-        result['missing_layers'] = missing
-        result['reachable_layers'] = reachable
-        if len(missing) > 0:
-            log.error('  Could not reach images:')
-            log.error('    ' + ', '.join(missing))
-            result['error'] = True
-            return result
+    # Testing for redirect
```
lcarva: Is this a leftover comment? Seems irrelevant here

bin/dock-pulp.py
```diff
@@ -214,139 +215,59 @@ def _test_repo(dpo, dockerid, redirect, pulp_imgs, protected=False, cert=None, k
...
 
-    # Testing for default pulp redirect, requires checking both v1 and v2 urls
-    else:
-        missingv1 = set([])
-        missingv2 = set([])
-        reachablev1 = set([])
-        reachablev2 = set([])
-        reponame = 'redhat-' + dockerid.replace('/', '-')
-        if dpo.url[:5] == 'https':
-            filerurl = dpo.url.replace('https', 'http', 1)
```
lcarva: Was the scheme check intentionally left out?

bin/dock-pulp.py
```diff
@@ -214,139 +215,59 @@ def _test_repo(dpo, dockerid, redirect, pulp_imgs, protected=False, cert=None, k
...
-            log.error('  Could not reach images:')
-            log.error('    ' + ', '.join(missing))
-            result['error'] = True
-            return result
+    # Testing for redirect
+    if not redirect:
+        reponame = 'redhat-' + dockerid.replace('/', '-')
+        redirect = dpo.url + '/pulp/docker/v1/' + reponame
+
+    missing = set([])
```
lcarva: `set([])` can be simplified to simply `set()`:

```
In [1]: set() == set([])
Out[1]: True
```

bin/dock-pulp.py
breillyrh: Yes, this is handled in /etc/dockpulp.conf

lcarva:
LGTM

PR #62, opened 2016-06-28
Prefix getter added to pulp object

breillyrh:


twaugh:
LGTM

twaugh:
Works fine here.

PR #64, opened 2016-06-30
Added history option to list

breillyrh:


bin/dock-pulp.py
```diff
@@ -821,19 +823,40 @@ def do_list(bopts, bargs):
                     except KeyError:
                         output[layer] = {}
 
-                    output[layer][manifest] = repo['manifests'][manifest]['tag']
+                    try:
```
lcarva: FYI, a more pythonic way of doing this:

```
output.setdefault('layer', {})
output['layer'].setdefault('manifest', {})
```

The above also creates the initial 'layer' key if needed (which is right above where it lets me comment)

dockpulp/__init__.py
```diff
@@ -743,6 +744,31 @@ def listRepos(self, repos=None, content=False):
                     r['manifests'][manifest['metadata']['digest']] = {}
                     r['manifests'][manifest['metadata']['digest']]['tag'] = manifest['metadata']['tag']
                     r['manifests'][manifest['metadata']['digest']]['layers'] = layers
+                if history:
+                    for manifest in r['manifests'].keys():
+                        data = self._get('/pulp/docker/v2/%s/manifests/%s' % (blob['id'], manifest))
+                        
+                        # Unsure if all v2 images will have v1 history
```
twaugh: Pretty sure they always will, but it's no bad thing to code defensively when handling docker things. :-)

twaugh:
LGTM

bin/dock-pulp.py
```diff
@@ -821,19 +823,40 @@ def do_list(bopts, bargs):
...
                 images = output.keys()
                 for image in images:
                     log.info('')
                     manifests = output[image].keys()
+                    tagoutput = []
                     for manifest in manifests:
-                        log.info('  Digest: %s  Tag: %s' %
-                            (manifest, output[image][manifest]))
-                    log.info('    Layers: ')
+                        log.info('  Manifest: %s  Tag: %s' %
```
lcarva: The preferred way is to pass each replacement as an arg to logging function:

```
log.info('  Manifest: %s  Tag: %s', manifest, output[image][manifest])
```

lcarva:
Tiny nitpicking comments, otherwise LGTM

bin/dock-pulp.py
breillyrh: Aha, that's much cleaner.

PR #68, opened 2016-07-11
Confirm does not test if there is no v1 or v2 content.

breillyrh:
Fixed issue with silent output not reporting v1 error if v2 content is fine as well.


twaugh:
LGTM

PR #73, opened 2016-07-18
sync: fix enable_v1 setting

twaugh:
The configuration option name is enable_v1 not enabled_v1:
https://github.com/pulp/pulp_docker/blob/master/docs/tech-reference/importer.rst

Reorganised code to avoid repetition.


breillyrh:
LGTM, merging.

PR #78, opened 2016-08-08
CLI: 'remove' should not calculate unneeded layers for hidden repo

twaugh:
I had the situation last week where I asked dock-pulp to remove a single layer from the hidden repository, and afterwards it decided to delete hundreds more layers.


breillyrh:
Ah, I don't think that's ever been tried before, as we would always keep all images in the hidden repo.

LGTM, merging.

PR #80, opened 2016-08-12
confirm: avoid traceback for unpublished v2 content

twaugh:
The check that the manifest's 'name' field matches the docker-id should only occur if the manifest actually exists.


breillyrh:
LGTM, merging.

PR #83, opened 2016-08-18
get_versions: skip layers if they don't have docker_version set

vrutkovs:
Recent docker versions seem to strip docker_version from all layers
except the last one during v1-to-v2 migration. We should ignore layers
that don't have docker_version key


twaugh:
Looks good to me.

- docker_version present but < 0.10? BAD
- docker_version present but >= 0.10? no problem
- docker_version not present? no problem, this is what newer docker parent layers look like now

breillyrh:
LGTM, merging.

PR #84, opened 2016-08-23
Added release order for distributors

breillyrh:


dockpulp/__init__.py
```diff
@@ -412,16 +413,23 @@ def crane(self, repos=[], wait=True, skip=False):
...
+                distributors.append(self.distributorconf[key])
+            for distributor in distributors:
+                dist_id = distributor['distributor_id']
+                override = {}
+                try:
+                    override = distributor['override_config']
+                except KeyError:
+                    pass
+
+                override['skip_fast_forward'] = skip
```
twaugh: Maybe it should only set this if skip is True? Otherwise there is no way to e.g. force it 'on' in the server configuration for the distributor.

Also this option is only valid for (I think) the cdn_distributor, so I wonder if it should only be added when the dist_id matches that id.

twaugh:
LGTM, minor suggestion about only setting skip_fast_forward if set.

PR #85, opened 2016-08-23
Mark tags as active if pullable

breillyrh:


dockpulp/__init__.py
```diff
@@ -781,6 +798,9 @@ def listRepos(self, repos=None, content=False, history=False):
                     r['manifests'][manifest['metadata']['digest']]['tag'] = manifest['metadata']['tag']
                     r['manifests'][manifest['metadata']['digest']]['layers'] = layers
 
+                for tag in tags:
+                    r['manifests'][tag['metadata']['manifest_digest']]['tag'] += ' (active)'
```
twaugh: This makes `--history` say things like:

```
4e3ba879ec3c4246048ec233d16c42177b049ef855ffbe24f4b3e29b788234bd (tags: v3.1.0.4 (active), v3.1.0.4-3 (active))
```
Maybe that's fine?

twaugh:
LGTM, thanks!

dockpulp/__init__.py
twaugh: Thinking more about this, I think the `--history` output should only include active tags, and not mark them specially. So the above output would instead look like:

```
4e3ba879ec3c4246048ec233d16c42177b049ef855ffbe24f4b3e29b788234bd (tags: v3.1.0.4, v3.1.0.4-3)
```

or, if no tags are active, `(tags: )`.

What do you think?

breillyrh: I think that's a good idea, I'll do that.

PR #86, opened 2016-08-31
Fix confirm errors caused by '(active)' changes

twaugh:
Instead of changing the tags that listRepos() reports, add a new 'tags' dict which maps tag names onto manifest digests.

In the CLI, use this new dict to decide which manifests have tags that are active.


vrutkovs:
Verified on a broken repo, +1

breillyrh:
LGTM, merging.

PR #88, opened 2016-09-01
Report missing v2 blobs

twaugh:
Previously listRepos() did not ask Pulp which blobs were present in the repository at all, and just assumed that any blobs referenced from the manifests were present.

This change alters that so that only the blobs which are actually present in Pulp are listed.

Blobs that are not present, but are referenced by a manifest, cause a warning to be emitted.

As part of refactoring this, I also combined the three separate Pulp API requests into a single request for speed/efficiency.


lcarva:
LGTM

breillyrh:
LGTM, merging.

PR #96, opened 2016-09-26
Fix E501 line too long (n > 100 characters)

vrutkovs:


dockpulp/__init__.py
```diff
@@ -1221,7 +1225,8 @@ def watch_tasks(self, task_ids, timeout=None, poll=5):
                 except AttributeError:
                     result = t.get("result", {}) if t.get("result", {}) else {}
                     reasons = result.get('reasons', [])
-                log.error(u"Pulp task [%s] failed:\nDetails:\n%s\nTags:\n%s\nReasons:\n%s\nException:\n%s\nTraceback:\n%s" %
+                log.error(u"Pulp task [%s] failed:\n" +
```
lcarva: Plus signed should be removed from end of this line. Otherwise, the string formatting on the next line will happen prior to string being concatenated, throwing a 'not all arguments converted' kind of error:

```python
>>> u'spam%s' + u'bacon%s' % ('yum', 'yum') 
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: not all arguments converted during string formatting
not all arguments converted during string formatting

>>> u'spam%s'  u'bacon%s' % ('yum', 'yum') 
u'spamyumbaconyum'
```

vrutkovs: Ouch, right. Fixing

bin/dock-pulp.py
```diff
@@ -598,7 +600,8 @@ def do_clone(bopts, bargs, parser):
         repoid = 'redhat-%s' % (args[1])
     else:
         if len(args) != 3:
-            parser.error('You need a source repo id, a new product line (rhel6, openshift3, etc), and a new image name')
+            parser.error('You need a source repo id, a new product line ' +
```
twaugh: I think '+' in general should be avoided when concatenating literal strings.

vrutkovs: Agreed, removed those

PR #97, opened 2016-09-26
content-url is now enforced to end with '/docker-id'

breillyrh:
Can be overridden with 'dock-pulp update' if needed.


lcarva:
Why is this change needed?

breillyrh:
This is a request from rel-eng.

twaugh:
Really? Is it possible to summarise the purpose of it? I don't think I have *ever* seen a Pulp repository whose redirect URL follows that pattern.

PR #98, opened 2016-09-29
Adding new tests for Pulp instance

lkolacek:


tests/test_pulp_instance.py
```diff
@@ -0,0 +1,56 @@
...
+        df.flush()
+        pulp = Pulp(env=name, config_file=fp.name, config_distributors=df.name)
+    return pulp
+
+
+# tests
+class TestPulp(object):
+    # Tests of methods from Pulp class.
+    def test_set_certs(self, pulp):
+        pulp.set_certs('foo', 'bar')
```
twaugh: Could use vars to hold the literal string values, to avoid having to repeat them. I'm being extra picky here because I know these tests will be used as templates for more tests in the future. :)

vrutkovs: +1, `pytest.mark.parametrize` will also help us to filter out failing cases

tests/test_pulp_instance.py
```diff
@@ -0,0 +1,56 @@
...
+        assert pulp.certificate == 'foo'
+        assert pulp.key == 'bar'
+
+    def test_getTask(self, pulp):
+        flexmock(RequestsHttpCaller)
+        RequestsHttpCaller.should_receive('__call__').with_args('get', '/pulp/api/v2/tasks/111/').once().and_return('well_done')
```
csomh: `RequestsHttpCaller.__call__()` seems to be super complicated, the idea was to restrict the scope of tests on `Pulp` and check parameters passed to `__call__`.
`RequestsHttpCaller` should have it's own tests, only focusing that class.

tests/test_pulp_instance.py
```diff
@@ -42,15 +42,19 @@ def pulp(tmpdir):
...
-
-    def test_getTask(self, pulp):
+    @pytest.mark.parametrize('cert, key', [('foo', 'bar')])
+    def test_set_certs(self, pulp, cert, key):
+        pulp.set_certs(cert, key)
+        assert pulp.certificate == cert
+        assert pulp.key == key
+
+    @pytest.mark.parametrize('tid, url, result', [
+        ('111', '/pulp/api/v2/tasks/111/', 'task_received')
```
twaugh: I think 'task_received' here isn't really part of what's being tested. It's more part of the testing machinery, so I think it ought to be declared inside the method rather than as a parameter.

lkolacek: Ok is this appropriate solution  http://pastebin.test.redhat.com/416622 ? Or should I use constants instead (for prefix '-redhat' etc.)?

twaugh: I'd prefer something like:
```
@pytest.mark.parametrize([...]
)
def [...]([...]):
    dummy_val = 'task_received'
    [...]
```

just as a guard against typing mistakes.

PR #101, opened 2016-10-18
Dockpulp checks update redirect-url to fit pulp standards

breillyrh:
This change is to resolve issue 2661.

Pulp fails silently on redirect-url updates if it does not fit a specific format, so dockpulp now front-ends that check for a better user experience.

### DIFF
--- a/tests/test_pulp_instance.py
+++ b/tests/test_pulp_instance.py
@@ -2,7 +2,7 @@
 # -*- coding: utf-8 -*-
 
 
-from dockpulp import Pulp, RequestsHttpCaller
+from dockpulp import Pulp, RequestsHttpCaller, errors
 import pytest
 from tempfile import NamedTemporaryFile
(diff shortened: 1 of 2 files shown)


lcarva:
Any chance we can add a corresponding unit test?

tests/test_pulp_instance.py
```diff
@@ -63,3 +63,19 @@ def test_getTask(self, pulp, tid, url):
 
     def test_getPrefix(self, pulp):
         assert pulp.getPrefix() == 'redhat-'
+
+    @pytest.mark.parametrize('rid, redirect', [
+        ('test-repo', 'http://example/url')
```
lcarva: would be nice to add more examples, like with `https://` and with multiple nested path: `http://example.com/spam/bacon/eggs/url`

could also add a negative test case to ensure bad values do raise DockPulpError

PR #102, opened 2016-10-19
Added listing label support for v2 manifests

breillyrh:
dockpulp list now displays label information for V2.

As far as I can tell, there is no way to obtain label information from Pulp for V1 images.


twaugh:
LGTM, nice.

PR #105, opened 2016-11-10
Added check-layers functionality

breillyrh:


dockpulp/__init__.py
```diff
@@ -327,6 +331,43 @@ def associate(self, dist_id, repo):
...
+        response['failedlayers'] = []
+        response['error'] = False
+        for img in images:
+            layer_url = '/pulp/docker/v1/%s/%s/layer' % (repo, img)
+            try:
+                r = self._get(layer_url, stream=True)
+                tar = tarfile.open(fileobj=r.raw, mode='r|*')
+                tar.close()
+            except (requests.exceptions.ConnectionError, IOError):
+                log.warning('Layer %s not available in pulp for repo %s' % (img, repo))
```
lcarva: `IOError` indicates it's in pulp, but file is corrupt no? Maybe different exception blocks so warnings can be more specific?

PR #107, opened 2016-11-18
Fixed createrepo enforcement

breillyrh:
Misunderstood initial requirement.

### DIFF
--- a/bin/dock-pulp.py
+++ b/bin/dock-pulp.py
@@ -783,6 +783,7 @@ def do_create(bopts, bargs, parser):
         elif (len(args) != 1 and len(args) != 2) and not p.isRedirect():
             parser.error('You need a name for a library-level repo')
         repoid = 'redhat-%s' % (args[0])
+        imagename = args[0]
         if len(args) == 2:
             url = args[1]
     else:
@@ -792,6 +793,7 @@ def do_create(bopts, bargs, parser):
         elif (len(args) != 2 and len(args) != 3) and not p.isRedirect():
             parser.error('You need a product line (rhel6, openshift3, etc) and image name')
         productid = args[0]
+        imagename = args[1]
         repoid = 'redhat-%s-%s' % (args[0], args[1])
         if len(args) == 3:
             url = args[2]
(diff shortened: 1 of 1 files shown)


bin/dock-pulp.py
```diff
@@ -792,15 +794,17 @@ def do_create(bopts, bargs, parser):
...
         if len(args) == 3:
             url = args[2]
 
     if url:
         if not url.startswith('/content'):
             parser.error('the content-url needs to start with /content')
-        if not url.endswith('/docker-id'):
-            parser.error('the content-url needs to end with /docker-id')
+        if not url.endswith("%s/%s" % (productid, imagename)) \
+           and not url.endswith("%s/%s/" % (productid, imagename)):
```
breillyrh: If it's 'or' it will fail on both <productid>/<imagename> and <productid>/<imagename>/

I agree with the stripping the trailing / though, working on that now...

bin/dock-pulp.py
```diff
@@ -792,15 +794,16 @@ def do_create(bopts, bargs, parser):
...
         repoid = 'redhat-%s-%s' % (args[0], args[1])
         if len(args) == 3:
             url = args[2]
 
     if url:
         if not url.startswith('/content'):
             parser.error('the content-url needs to start with /content')
-        if not url.endswith('/docker-id'):
-            parser.error('the content-url needs to end with /docker-id')
+        if not url.rstrip('/').endswith("%s/%s" % (productid, imagename)):
```
lcarva: Will this work with empty `productid`? Or is that not a valid scenario?

lcarva:
Could add tests?

bin/dock-pulp.py
breillyrh: An empty `productid` is only valid if the `--library` option is set.
I've tested that scenario, it works as expected (where `productid` is set to '')

breillyrh:
This is all CLI code, can't really add tests until I refactor it.

bin/dock-pulp.py
lcarva: great, thanks for clarifying. new question:

This check will not ensure that the productid is a full match, only that it endswith it. Should it be more strict?

bin/dock-pulp.py
```diff
@@ -792,15 +793,19 @@ def do_create(bopts, bargs, parser):
...
         repoid = 'redhat-%s-%s' % (args[0], args[1])
         if len(args) == 3:
             url = args[2]
 
     if url:
         if not url.startswith('/content'):
             parser.error('the content-url needs to start with /content')
-        if not url.endswith('/docker-id'):
-            parser.error('the content-url needs to end with /docker-id')
+        if productid:
```
lcarva: Would be nice if conditional only applied to determining the expected suffix:

```python
suffix = '/%s' % imagename
if productid:
   suffix = '/%s%s' % (productid, suffix)
```

PR #109, opened 2016-12-14
Added silent option for json dump of repo listing

breillyrh:


breillyrh:
JSON format was specifically asked for by rel-eng - they are also the only consumers of the CLI code at the moment. You can mimic this functionality by calling listRepos directly... so I am unsure of what could potentially change here that would break processes.

PR #110, opened 2016-12-21
Moved CLI code to separate file to allow for testing

breillyrh:


dockpulp/cli.py
```diff
@@ -0,0 +1,941 @@
...
+formatter = logging.Formatter("%(levelname)-9s %(message)s")
+sh.setFormatter(formatter)
+log.addHandler(sh)
+
+silent = logging.StreamHandler(sys.stdout)
+silformatter = logging.Formatter("")
+silent.setFormatter(silformatter)
+
+
+def main():
```
lcarva: Let's allow this method to take an optional list of args to facilitate testing.

```
def main(args=None):
  ...
  opts, args = parser.parse_args(args)
  ...
```

Then, if called as `main()`, it'll use sys.argv (https://hg.python.org/cpython/file/2.7/Lib/optparse.py#l1368)
Otherwise, `main(['foo', 'bar'])` allows overriding sys.argv with the list `['foo', 'bar']`.

PR #112, opened 2017-01-11
Added unit test for pulp login

breillyrh:


breillyrh:
Just realized this one needs a revision... will break out into its own test "CLI" class.

tests/test_pulp_instance.py
```diff
@@ -268,3 +289,19 @@ def test_checkBlobsFail(self, pulp, repo, blob):
             .and_raise(requests.exceptions.ConnectionError))
         response = pulp.checkBlobs(repo, blobs)
         assert response['error']
+
+# cli tests
+    def test_pulplogin(self):
```
lcarva: Maybe a new `test_cli.py` file?

lcarva: Would be great to test the different code paths in `pulp_login` method.

tests/test_pulp_instance.py
```diff
@@ -268,3 +289,19 @@ def test_checkBlobsFail(self, pulp, repo, blob):
...
+        bopts = testbOpts()
+        p = testPulp()
+        flexmock(Pulp)
+        (Pulp
+            .should_receive('__new__')
+            .once()
+            .and_return(p))
+        flexmock(os.path)
+        (os.path
```
lcarva: There's a way to use flexmock to specify the args that should have been used for each of the calls.

tests/test_cli.py
```diff
@@ -0,0 +1,68 @@
...
+
+# tests
+class TestCLI(object):
+    # Tests of methods from CLI
+    @pytest.mark.parametrize('server, config_file, debug, cert, key, error',
+                             [('testserv', 'testconf', None, None, None, False),
+                              ('testserv', 'testconf', True, None, None, False),
+                              ('testserv', 'testconf', None, True, True, False),
+                              ('testserv', 'testconf', None, None, None, True)])
+    def test_pulplogin(self, server, config_file, debug, cert, key, error):
```
lcarva: Should probably be `test_pulp_login`.

tests/test_cli.py
```diff
@@ -0,0 +1,68 @@
...
+        return
+
+    def setDebug():
+        return
+
+
+# tests
+class TestCLI(object):
+    # Tests of methods from CLI
+    @pytest.mark.parametrize('server, config_file, debug, cert, key, error',
```
lcarva: Since `server` and `config_file` never change, I suggest just making local variables in the `test_pulp_login` method.

Also, `debug` seems unrelated to `cert`, `key`. And `error` seems unrelated as well. You could have multiple `parametrize` decorators:

```python
    @pytest.mark.parametrize('debug', [True, False])
    @pytest.mark.parametrize('error', [True, False])
    @pytest.mark.parametrize('cert, key', [...])
    def test_pulplogin(self, debug, cert, key, error):
```

tests/test_cli.py
```diff
@@ -0,0 +1,66 @@
...
+# tests
+class TestCLI(object):
+    # Tests of methods from CLI
+    @pytest.mark.parametrize('debug', [True, False])
+    @pytest.mark.parametrize('error', [True, False])
+    @pytest.mark.parametrize('cert, key', [(None, True), (True, None), (True, True), (None, None)])
+    def test_pulp_login(self, debug, cert, key, error):
+        bopts = testbOpts("testserv", "testconf", debug, cert, key)
+        p = testPulp()
+        flexmock(Pulp).new_instances(p) \
```
twaugh: Recommended way of splitting this over multiple lines is like this:
```
(flexmock(Pulp)
    .new_instances(p)
    .with_args(...))
```

tests/test_cli.py
```diff
@@ -0,0 +1,66 @@
...
+            .with_args(Pulp, env=bopts.server, config_file=bopts.config_file)
+        if debug:
+            flexmock(testPulp)
+            (testPulp
+                .should_receive('setDebug')
+                .once()
+                .and_return(None))
+        flexmock(os.path)
+        if error and not (key or cert):
+            (os.path
```
twaugh: This looks odd. Shouldn't this be the `Expectation` that flexmock(os.path) returns?

PR #114, opened 2017-01-18
Make tests a module to aid local coverage testing

twaugh:
Doing this makes 'py.test' work locally, which also means you can install pytest-cov and use 'py.test --cov dockpulp' to get a coverage report. You can also use this to get it annotated by line: 'py.test --cov dockpulp --cov-report=annotate'


twaugh:
I couldn't get tox to give a coverage report.

PR #116, opened 2017-01-19
Added unit test for clone

breillyrh:


tests/test_cli.py
```diff
@@ -110,6 +119,49 @@ def test_do_associate(self, bargs):
...
+        flexmock(testPulp)
+        (testPulp
+            .should_receive('listRepos')
+            .and_return(oldinfo))
+        (testPulp
+            .should_receive('createRepo')
+            .and_return(None))
+        if img:
+            (testPulp
+                .should_receive('copy')
```
twaugh: `.once()`?

tests/test_cli.py
```diff
@@ -110,6 +119,49 @@ def test_do_associate(self, bargs):
...
+                .and_return(None))
+            (testPulp
+                .should_receive('updateRepo')
```
twaugh: `.once()`? Also, do we need to check the parameters to these calls?

PR #117, opened 2017-02-13
Added option to set signatures to sign repos

breillyr:


conf/dockpulp.conf
```diff
@@ -61,3 +61,8 @@ test =
 #prod = 1
 #stage = 1
 #test = 1
+
+# Signature keys allowed when creating signed repos
+# This section is optional and environment independent
+#[signatures]
+#beta = foobar
```
lcarva: Would be nice to have a new line here

dockpulp/__init__.py
```diff
@@ -1118,6 +1125,20 @@ def getRepos(self, rids, fields=None):
         return self._post('/pulp/api/v2/repositories/search/',
                           data=json.dumps(data))
 
+    def getSignature(self, sig):
+        """Return a signature key."""
+        try:
+            self.sigs[sig]
+        except AttributeError:
```
twaugh: Could just handle this exception in the `return...` block below?

breillyr: I want to print out the available keys as part of the exception catch - if self.sigs doesn't exist that's going to cause an attribute error at the print.

dockpulp/__init__.py
```diff
@@ -1118,6 +1125,20 @@ def getRepos(self, rids, fields=None):
...
+            log.error('Signatures not defined in dockpulp.conf')
+            exit(1)
```
lcarva: How about raising an exception instead of quitting the process?

dockpulp/__init__.py
twaugh: I just wondered why the code wasn't written like:
```
try:
    return self.sigs[sig]
except AttributeError:
    # handle AttributeError...
except KeyError:
    # handle KeyError...
```

breillyr: Oh! I wasn't actually aware you could write try-except blocks that way...good to know for the future.

PR #118, opened 2017-02-23
Fixed library docker-id bug, added unit tests for createRepo

breillyr:


tests/test_pulp_instance.py
```diff
@@ -268,3 +274,24 @@ def test_checkBlobsFail(self, pulp, repo, blob):
...
+    @pytest.mark.parametrize('url', [None, 'http://test', '/content/foo/bar'])
+    @pytest.mark.parametrize('registry_id', [None, 'foo/bar'])
+    @pytest.mark.parametrize('sig', [None, 'foobar'])
+    @pytest.mark.parametrize('distributors', [True, False])
+    @pytest.mark.parametrize('library', [True, False])
+    def test_createRepo(self, pulp, repo_id, url, registry_id, sig, distributors, productline,
+                        library):
+        flexmock(RequestsHttpCaller)
+        (RequestsHttpCaller
+            .should_receive('__call__')
```
twaugh: Do we know what args we expect that call to be given?

breillyr: IMO it's too complex to figure out exactly what's being sent - the whole point of createRepo is to construct the data block, which is one of the arguments sent. I'd basically be duplicating the createRepo code in the test just to check.

I have added more assertions to check those values that we care about.

PR #119, opened 2017-02-24
Added unit tests for associate / disassociate

breillyr:


tests/test_pulp_instance.py
```diff
@@ -278,6 +278,20 @@ def test_checkBlobsFail(self, pulp, repo, blob):
         response = pulp.checkBlobs(repo, blobs)
         assert response['error']
 
+    @pytest.mark.parametrize('dist_id', ['foo', 'bar'])
+    def test_associate(self, pulp, dist_id):
+        if dist_id == 'bar':
+            with pytest.raises(SystemExit):
```
twaugh: Why `SystemExit`? Shouldn't this be `KeyError`?

Also do we expect `RequestsHttpCaller.__call__` to be called at all for this?

tests/test_pulp_instance.py
```diff
@@ -318,3 +332,22 @@ def test_createRepo(self, pulp, repo_id, url, registry_id, sig, distributors, pr
...
+        dist_id = 'foo'
+        dl = '/pulp/api/v2/repositories/%s/distributors/%s/' % (repo, dist_id)
+        tid = '123'
+        url = '/pulp/api/v2/tasks/%s/' % tid
+        t = {'state': 'finished'}
+        flexmock(RequestsHttpCaller)
+        (RequestsHttpCaller
+            .should_receive('__call__')
+            .with_args('delete', dl)
+            .and_return(tid))
```
twaugh: `.once()`?

tests/test_pulp_instance.py
```diff
@@ -318,3 +332,22 @@ def test_createRepo(self, pulp, repo_id, url, registry_id, sig, distributors, pr
...
+        (RequestsHttpCaller
+            .should_receive('__call__')
+            .with_args('get', url)
+            .and_return(t))
```
twaugh: `.once()`?

tests/test_pulp_instance.py
breillyr: The KeyError is excepted in the code so it wouldn't be what we're looking for here.
I did however decide to change the systemexit to a configuration error, to make it more clear
to the end user what the problem is.

We do not expect that call for this check - that's specifically the pulp '_delete' which is never reached if the distributor id is incorrect.

twaugh: OK.

Could add a
```
(flexmock(RequestsHttpCaller)
    .should_receive('__call__')
    .never())
```

PR #120, opened 2017-03-10
Added option for distribution field in repos

breillyr:


conf/dockpulp.conf
```diff
@@ -66,3 +66,8 @@ test =
 # This section is optional and environment independent
 #[signatures]
 #beta = foobar
+
+# Distributions allowed when creating repos in a comma separated list
+# This section is optional and environment independent
+#[distribution]
+#distributions = beta,ga
```
lcarva: Ending a file in a new line is usually good practice.

dockpulp/__init__.py
```diff
@@ -651,6 +652,10 @@ def _set_cert(self, attrs):
                 self.key = os.path.join(os.path.expanduser(cert_path),
                                         self.AUTH_KEY_FILE)
 
+    def _set_dist_attr(self, attrs):
+        for key, val in attrs:
+            return val
```
lcarva: Don't we need the env check here? `if self.env == key`

dockpulp/cli.py
```diff
@@ -229,10 +229,21 @@ def do_clone(bopts, bargs, parser):
 
     log.info('cloning %s repo to %s' % (args[0], repoid))
     oldinfo = p.listRepos(args[0], content=True)[0]
+
+    try:
```
lcarva: Could simplify both of these: 

```python
sig = oldinfo.get('signatures')
dist = oldinfo.get('distribution')
```

dockpulp/__init__.py
breillyr: Nope, this is environment independent

twaugh: So the key name is deliberately ignored? Could add a comment just pointing out that the first value is used, regardless of key name (to show this is deliberate).

dockpulp/__init__.py
```diff
@@ -831,6 +836,18 @@ def checkBlobs(self, repo, blobs):
                 response['error'] = True
         return response
 
+    def checkDistribution(self, dist):
+        """Check if distribution is valid."""
+        try:
+            if dist in self.dists.split(','):
+                return
+            else:
```
twaugh: The 'else' path isn't covered by testing.

dockpulp/__init__.py
```diff
@@ -831,6 +836,18 @@ def checkBlobs(self, repo, blobs):
...
+                log.error('Distribution not defined in dockpulp.conf')
+                raise errors.DockPulpConfigError(
+                    'Available distributions are: %s' % self.dists)
+        except AttributeError:
```
twaugh: The 'except' path isn't covered by testing.

dockpulp/__init__.py
```diff
@@ -1629,6 +1654,9 @@ def updateRepo(self, rid, update):
         if 'signature' in update:
             sig = self.getSignature(update['signature'])
             delta['delta']['notes'] = {'signatures': sig}
+        if 'distribution' in update:
```
twaugh: The 'if' path isn't covered by testing.

PR #121, opened 2017-03-22
Clone now copies V2 images

breillyr:


dockpulp/cli.py
```diff
@@ -241,7 +241,10 @@ def do_clone(bopts, bargs, parser):
             p.copy(repoid, img)
             tags = {'tag': '%s:%s' % (','.join(oldinfo['images'][img]), img)}
             p.updateRepo(repoid, tags)
-    else:
+    if len(oldinfo['manifests']) > 0:
```
lcarva: Is copying manifests sufficient to trigger copy of blobs and tags?

PR #122, opened 2017-03-24
Added noprefix option for create and clone

breillyr:


dockpulp/cli.py
```diff
@@ -213,28 +213,39 @@ def do_clone(bopts, bargs, parser):
...
+        else:
+            repoid = 'redhat-%s-%s' % (args[1], args[2])
         productid = args[1]
 
     log.info('cloning %s repo to %s' % (args[0], repoid))
     oldinfo = p.listRepos(args[0], content=True)[0]
     sig = oldinfo.get('signatures')
     dist = oldinfo.get('distribution')
+    if opts.noprefix:
+        prefix_with = ''
```
lcarva: You could set this earlier in the method to avoid the additional conditionals:

```python
prefix_with = '' if opts.noprefix else 'redhat-'
if opts.library:
    ...
    repoid = '%s%s' % (prefix_with, args[1])
    ...
else:
    ...
    repoid = '%s%s-%s' % (prefix_with, args[1], args[2])
    ...

```

Same for `do_create`.

PR #123, opened 2017-03-31
Make source and installation filenames consistent

rohanpm:
Currently, these files end in .py in the source tree and in setup.py,
and don't when installed from RPM.

It's good to avoid this inconsistency, so that the command is
"dock-pulp" regardless of whether the project was installed from RPM or
into a virtualenv.


breillyr:
LGTM, thanks for the patch

PR #124, opened 2017-04-05
Made dock-pulp-bootstrap idempotent, added sigstore repo setup

breillyr:


tests/test_pulp_instance.py
```diff
@@ -339,22 +339,24 @@ def test_associate(self, pulp, dist_id):
                                                       ('foo-bar', 'foo'),
                                                       ('foo', None),
                                                       ('foo-bar-test', 'foo-bar')])
+    @pytest.mark.parametrize('repotype, importer_type_id', [('foo', 'bar'), (None, None)])
```
breillyr: Tried to reduce some of the computational complexity here... both variables are independent of one another and fairly inconsequential (only used by dock-pulp-bootstrap).

PR #126, opened 2017-04-18
Added missing error return

breillyr:


dockpulp/__init__.py
```diff
@@ -288,6 +288,7 @@ def _test_repo(self, dockerid, redirect, pulp_imgs, protected=False, silent=Fals
             log.error('  Pulp images and Crane images are not the same:')
             if pdiff:
                 log.error('    In Pulp but not Crane: ' + pdiff)
+                result['error'] = True
```
twaugh: Could move this (and the one below) out to the `if pdiff or cdiff` block.

PR #127, opened 2017-04-25
Added distribution mapping to signatures

breillyr:


dockpulp/__init__.py
```diff
@@ -839,17 +837,17 @@ def checkBlobs(self, repo, blobs):
...
-        """Check if distribution is valid."""
+    def getDistributionSig(self, dist):
+        """Check distribution and get its associated signature."""
         try:
-            if dist in self.dists.split(','):
-                return
-            else:
+            return self.dists[dist]
+        except (KeyError, AttributeError) as e:
+            if type(e) is KeyError:
```
twaugh: Maybe `except KeyError:` should be separated from `except AttributeError:`?

tests/test_pulp_instance.py
```diff
@@ -372,8 +372,6 @@ def test_createRepo(self, pulp, repo_id, url, registry_id, sig, distributors, pr
             rurl = pulp.cdnhost + url
         assert response['id'] == repo_id
         assert response['display_name'] == repo_id
-        if sig:
-            assert response['notes']['signatures'] == pulp.getSignature(sig)
```
twaugh: Don't we still want to make some assertion about notes.signatures?

PR #128, opened 2017-05-08
Added configurable naming enforcement for repo-ids and content-urls

breillyr:


conf/dockpulp.conf
```diff
@@ -72,3 +72,13 @@ test =
...
+
+# Repo-id name enforcement to reflect distribution (must end repo name with -<distname>)
+# This section is optional and environment independent
+#[name_enforce]
+#beta = yes
+
+# Content-url enforcement to reflect distribution (url must start with what is specified)
+# This section is optional and environment independent
+#[content_enforce]
+#beta = /content/beta
```
lcarva: It's good practice to end files with a newline char.

breillyr:
Getting a strange flake8 traceback from the CI but tox works locally...not sure how I can fix that

dockpulp/__init__.py
```diff
@@ -1010,6 +1020,19 @@ def createRepo(self, repo_id, url, registry_id=None, desc=None, title=None, prot
             'notes': {'_repo-type': 'docker-repo'},
         }
         if distribution:
+            try:
+                if not repo_id.endswith(self.name_enforce.get(distribution, '')):
+                    raise errors.DockPulpError("%s is a %s repo, repo id must end with %s" %
+                                               (repo_id, distribution, distribution))
+            except AttributeError:
```
lcarva: Is this to handle the case where `name_enforce` is not used? If so, maybe it's worth initializing it to an empty dict?

dockpulp/__init__.py
```diff
@@ -1010,6 +1021,16 @@ def createRepo(self, repo_id, url, registry_id=None, desc=None, title=None, prot
...
         if distribution:
+            if not repo_id.endswith(self.name_enforce.get(distribution, '')):
+                raise errors.DockPulpError("%s is a %s repo, repo id must end with %s" %
+                                           (repo_id, distribution, distribution))
+            try:
+                if not url.startswith(self.content_enforce.get(distribution, '')):
+                    raise errors.DockPulpError("%s is a %s repo, content-url must start with %s" %
+                                               (repo_id, distribution,
+                                                self.content_enforce[distribution]))
+            except AttributeError:
```
lcarva: Same comment regarding setting default value here.

breillyr: In this one case the AttributeError covers both the case where the dict doesn't exist, and where there is no url defined, since a content-url is not a requirement for every environment.

PR #130, opened 2017-05-22
Refactored distribution conf into json

breillyr:


dockpulp/__init__.py
```diff
@@ -1021,25 +1015,38 @@ def createRepo(self, repo_id, url, registry_id=None, desc=None, title=None, prot
...
+                raise errors.DockPulpConfigError("Distribution %s not defined in %s" %
+                                                 (distribution, DEFAULT_DISTRIBUTIONS_FILE))
             if productline:
-                if not productline.endswith(self.name_enforce.get(distribution, '')):
-                    raise errors.DockPulpError("%s is a %s repo, product-line must end with %s" %
+                if not productline.endswith(distconf.get('name_enforce', '')):
+                    raise errors.DockPulpError("%s is a %s repo, product-line must end with -%s" %
                                                (repo_id, distribution, distribution))
+                for restrict in distconf['name_restrict']:
+                    if productline.find(restrict) != -1:
```
lcarva: Would it be cleaner to simply: `if restrict in productline:` ?

dockpulp/__init__.py
```diff
@@ -1021,25 +1015,38 @@ def createRepo(self, repo_id, url, registry_id=None, desc=None, title=None, prot
...
+                            "%s is a %s repo, product-line must not contain %s" %
+                            (repo_id, distribution, restrict))
             # library level repo do not use product-line
             elif library:
-                if not repo_id.endswith(self.name_enforce.get(distribution, '')):
+                if not repo_id.endswith(distconf.get('name_enforce', '')):
                     raise errors.DockPulpError("%s is a %s repo, repo id must end with %s" %
                                                (repo_id, distribution, distribution))
+                for restrict in distconf['name_restrict']:
+                    if repo_id.find(restrict) != -1:
```
lcarva: Would it be cleaner to simply: `if restrict in repo_id:` ?

dockpulp/__init__.py
```diff
@@ -1021,25 +1015,38 @@ def createRepo(self, repo_id, url, registry_id=None, desc=None, title=None, prot
...
+            try:
+                distconf = self.distributionconf[distribution]
+            except KeyError:
+                raise errors.DockPulpConfigError("Distribution %s not defined in %s" %
+                                                 (distribution, DEFAULT_DISTRIBUTIONS_FILE))
             if productline:
-                if not productline.endswith(self.name_enforce.get(distribution, '')):
-                    raise errors.DockPulpError("%s is a %s repo, product-line must end with %s" %
+                if not productline.endswith(distconf.get('name_enforce', '')):
+                    raise errors.DockPulpError("%s is a %s repo, product-line must end with -%s" %
```
lcarva: Where's the dash, `-`, being enforced? Should the check be if productline endswith `'-' + name_enforce`? Or should the dash be removed from error message?

breillyr: name_enforce is now a string, so the dash can be explicitly defined in the json

breillyr: Ah but I see your point now :) Will remove it from the error message...

PR #134, opened 2017-07-05
Fixed bug with distribution and hidden repos

breillyr:


twaugh:
Can this be part of the automated tests?

PR #137, opened 2017-07-17
New pulp (2.13) doesn't store tags along with manifest, so it should be looked up additionally

vrutkovs:
PTAL @breillyr @lcarva @rcerven


breillyr:
LGTM.

PR #138, opened 2017-07-21
Fixed bug with cloning repos with distribution defined and rurl enfor…

breillyr:
…cement

### DIFF
--- a/dockpulp/__init__.py
+++ b/dockpulp/__init__.py
@@ -29,6 +29,7 @@ import time
 import warnings
 from contextlib import closing
 from functools import wraps
+from urlparse import urlparse
 
 import multiprocessing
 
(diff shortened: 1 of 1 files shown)


dockpulp/__init__.py
```diff
@@ -1029,6 +1029,9 @@ def createRepo(self, repo_id, url, registry_id=None, desc=None, title=None, prot
                 registry_id = repo_id.replace(prefix_with, '').replace('-', '/', 1)
 
         rurl = url
+        if url and url.startswith('http'):
+            # want to strip off hostname info here
+            url = url[url.find('/', 10):]
```
lcarva: Why `10` ?

breillyr: Just to get past 'http://' or 'https://' to find the next '/'

lcarva: How about using urlparse? https://docs.python.org/2/library/urlparse.html#urlparse.urlparse

PR #140, opened 2017-08-01
Changed request retry implementation

breillyr:
I'm not sure how to mimic connection failures with flexmock in such a way for the adapter to handle them so I've commented out the relevant unit tests for now.

### DIFF
--- a/dockpulp/__init__.py
+++ b/dockpulp/__init__.py
@@ -28,9 +28,9 @@ import tempfile
 import time
 import warnings
 from contextlib import closing
-from functools import wraps
 from urlparse import urlparse
-
+from requests.adapters import HTTPAdapter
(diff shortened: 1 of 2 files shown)


twaugh:
Here's one way of generating HTTP failures on demand:
https://github.com/projectatomic/osbs-client/pull/639/files#diff-6803485e05312ed68eeb5295e8517a06R178

twaugh:
On seconds thoughts, we don't really want to have automated tests take 15 minutes apiece as well as being dependent on httpbin.org for that time.

Related pull request: https://github.com/projectatomic/osbs-client/pull/657

breillyr:
I've updated the tests to a similar implementation as OSBS.

PR #141, opened 2017-08-18
Improved sigstore repo checking

breillyr:
Just an initial patchset - I will be updating the PR with unit tests.

### DIFF
--- a/dockpulp/__init__.py
+++ b/dockpulp/__init__.py
@@ -27,6 +27,12 @@ import tarfile
 import tempfile
 import time
 import warnings
+try:
+    # Want to use gnupg here but there are issues with python 2.6
+    import gnupg
+except ImportError:
+    gnupg = None
+    import subprocess
 from contextlib import closing
 from urlparse import urlparse
 from requests.adapters import HTTPAdapter
@@ -157,6 +163,18 @@ class Crane(object):
         self.p = pulp
         self.cert = cert
         self.key = key
+        self.requests = RequestsHttpCaller(None, pulp.retries).requests_retry_session()
+
+    def _split_signature(self, signature, prefix_with):
+        # splits signatures from sigstore and returns (repo, manifest)
+        # signatures are in the form "productline/imagename@sha256=shasum/signature-1"
+        (repo, manifest) = signature.split('@')
+        if not repo.startswith(prefix_with):
+            repo = prefix_with + repo
+        repo = repo.replace('/', '-')
+        manifest = manifest.split('/')[0]
+        manifest = manifest.replace('=', ':')
+        return (repo, manifest)
 
     def confirm(self, repos, v1=True, v2=True, silent=True, check_layers=False):
 
@@ -175,7 +193,7 @@ class Crane(object):
             repoids[repo['id']] = {}
             errorids[repo['id']] = False
             if repo['id'] == SIGSTORE:
-                response = self._test_sigstore(repo['sigstore'])
+                response = self._test_sigstore(repo['sigstore'], exception=self.p.sig_exception)
                 if silent:
                     repoids[repo['id']].update(response)
                     if response['error']:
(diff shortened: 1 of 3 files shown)


dockpulp/__init__.py
```diff
@@ -610,25 +617,75 @@ def _test_repoV2(self, dockerid, redirect, pulp_manifests, pulp_blobs, pulp_tags
...
             log.debug('  status code: %s', answer.status_code)
             if not answer.ok:
                 log.error('  Signature missing in CDN: %s', signature)
                 result['error'] = True
                 result['sigs_in_pulp_not_crane'].append(signature)
+            if gnupg:
+                data = gpg.decrypt(answer.content)
+                key_id = data.key_id
+            else:
+                data = subprocess.Popen(['gpg', '-d'], stdin=subprocess.PIPE,
```
bfontecc007: What if the system doesn't have gpg?

breillyr: A good question, but I am not entirely sure if I want to make dockpulp have gpg as a build dependency. I suppose if gpg is not installed this check can be bypassed entirely.

dockpulp/__init__.py
```diff
@@ -610,25 +617,75 @@ def _test_repoV2(self, dockerid, redirect, pulp_manifests, pulp_blobs, pulp_tags
...
+            repo = repo.replace('/', '-')
+            manifest = manifest.split('/')[0]
+            manifest = manifest.replace('=', ':')
+            manifests.setdefault(repo, []).append(manifest)
+        signed_repos = self.p.listRepos(manifests.keys(), content=True, strict=False)
+        repo_sigs = {}
+        for repo in signed_repos:
+            repo_sigs[repo['id']] = repo['signatures']
+            sigstoremanifests = set(manifests[repo['id']])
+            if not set(repo['manifests'].keys()).issuperset(sigstoremanifests):
```
lcarva: Minor simplification would be to always compute `difference` and just use `mdiff` value here.

dockpulp/__init__.py
```diff
@@ -610,25 +617,75 @@ def _test_repoV2(self, dockerid, redirect, pulp_manifests, pulp_blobs, pulp_tags
...
+                mdiff = sigstoremanifests.difference(set(repo['manifests'].keys()))
+                for m in list(mdiff):
+                    log.error('  Signatures in sigstore but not in repo %s:', repo['id'])
```
lcarva: `m` appears to be just a single signature. Do we want to output this line for every one? or just once per repo?
Or maybe just move it outside of the for-block?

dockpulp/__init__.py
```diff
@@ -610,25 +617,75 @@ def _test_repoV2(self, dockerid, redirect, pulp_manifests, pulp_blobs, pulp_tags
...
+                    log.error('    %s', m)
+                    result['manifests_in_sigstore_not_repo'].append(m)
+                result['error'] = True
         url = self.p.cdnhost + '/content/sigstore/'
-        log.info('  Confirming CDN has signatures available')
+        log.info('  Confirming CDN has valid signatures available')
+        if gnupg:
+            gpg = gnupg.GPG()
         for signature in signatures:
+            (repo, manifest) = signature.split('@')
```
lcarva: There's some code duplication here. Maybe a helper method to return `repo` and `manifest` from a signature?

dockpulp/__init__.py
```diff
@@ -610,25 +617,75 @@ def _test_repoV2(self, dockerid, redirect, pulp_manifests, pulp_blobs, pulp_tags
...
+                for m in list(mdiff):
+                    log.error('  Signatures in sigstore but not in repo %s:', repo['id'])
+                    log.error('    %s', m)
+                    result['manifests_in_sigstore_not_repo'].append(m)
+                result['error'] = True
         url = self.p.cdnhost + '/content/sigstore/'
-        log.info('  Confirming CDN has signatures available')
+        log.info('  Confirming CDN has valid signatures available')
+        if gnupg:
+            gpg = gnupg.GPG()
```
lcarva: It might make more sense to instantiate this class when it's actually used.

dockpulp/__init__.py
bfontecc007: Might want to add a status check to see if Popen was able to run gpg -d.

lcarva: wouldn't it raise an exception if return code is not 0?

bfontecc007: *Googles it

Yes, looks like it should raise CalledProcessError

bfontecc007: Might be worth trapping CalledProcessError and throwing something telling the use to install gpg to their system. Personally, I really get frustrated having to figure out what packages are needed by interpreting all different tracebacks. Just my two cents. If I were breilly I'd probably add the dep, if possible.

twaugh: Related question: what about gpg vs gpg2?

dockpulp/__init__.py
```diff
@@ -157,6 +163,17 @@ def __init__(self, pulp, cert=None, key=None):
...
+        self.requests = RequestsHttpCaller(None, pulp.retries).requests_retry_session()
+
+    def _split_signature(self, signature, prefix_with):
+        # splits signatures from sigstore and returns (repo, manifest)
+        (repo, manifest) = signature.split('@')
+        if not repo.startswith(prefix_with):
+            repo = prefix_with + repo
+        repo = repo.replace('/', '-')
+        manifest = manifest.split('/')[0]
+        manifest = manifest.replace('=', ':')
```
twaugh: I don't think I understand the format of 'signature', and I can't tell if we're testing all these parts. What is the '=' for?

tests/test_pulp_instance.py
```diff
@@ -582,21 +611,33 @@ def test_test_sigstore(self, crane, signatures):
...
             .should_receive('get')
             .with_args(url + 'PULP_MANIFEST', verify=False)
             .once()
             .and_return(answer2))
         response = crane._test_sigstore(signatures)
-        if signature == 'failedsig':
-            assert response == {'error': True, 'sigs_in_pulp_not_crane': ['failedsig'],
-                                'sigs_in_crane_not_pulp': ['sig@shasum']}
-        elif signature == 'failedcrane':
+        if signature == 'failed1@sig':
```
twaugh: To avoid all the `if` cases for different signatures, how about parametrizing how they behave?
```
@pytest.mark.parametrize('signature,response_ok,sig_stderr,expected_result', [ ... ]
```

dockpulp/__init__.py
breillyr: Signatures come in the format "<productline>/<imagename>@sha256=<shasum>/signature-1"

Swapping out the '=' for a ':' gives us the equivalent manifest for the signature.

twaugh: OK. Might be good to include that in a comment alongside the code.

PR #142, opened 2017-08-22
Separate credentials by server

ahills:
It's extremely annoying to have to log in each time I switch servers. Reconciling problems efficiently between environments (especially prod and stage) results in sensitive passwords being stored in plaintext in scripts and aliases. This simple patch uses the environment name to distinguish between credential storage files.

I opted to keep the files the same name in the session directory, rather than changing the filename in `do_login()`. I don't have a preference for one method over the other, though.

### DIFF
--- a/dockpulp/__init__.py
+++ b/dockpulp/__init__.py
@@ -657,8 +657,8 @@ class Pulp(object):
                               ('retries', "_set_int_attr", "retries"),
                               ('distribution', "_set_bool", "dists"),
                               ('signatures', "_set_independent_attr", "sigs"))
-    AUTH_CER_FILE = "pulp.cer"
-    AUTH_KEY_FILE = "pulp.key"
+    AUTH_CER_FILE = "pulp-%s.cer"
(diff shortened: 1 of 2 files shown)


lcarva:
My only concern is backward compatibility. Some of our tooling, i.e. atomic-reactor, that I believe assume the non env file name. 
To be safe, when reading the cert/key file, can we fallback to the old one in that case?

twaugh:
atomic-reactor uses `set_certs` to tell dockpulp which files to use: https://github.com/release-engineering/dockpulp/blob/dockpulp-1.43-1/dockpulp/__init__.py#L1621-L1624

But it is a good idea anyway to fall back to the old names if the new names cannot be opened.

lcarva:
@ahills, do you plan on addressing the requested changes?

pbabinca:
I have rebased this PR and fixed current tests. I didn't know how to alter this PR so have created new one -#196.

breillyr:
This PR has been duplicated and merged, closing this one.

PR #146, opened 2017-09-20
Improved sync with filter copy

breillyr:


dockpulp/__init__.py
```diff
@@ -1766,11 +1766,13 @@ def syncRepo(self, env=None, repo=None, config_file=DEFAULT_CONFIG_FILE,
         manifests.sort()
 
         # Need to maintain HIDDEN
+        pulp_filter = {'unit': {'$and': []}}
```
twaugh: Don't need to set this outside of the `if` block.

We should avoid repeating 'unit' and '$and'.

Is '$and' really correct? I can't imagine this filter would match anything:
```
{
  "unit": {
    "$and": {
      [
        {
          "digest": "something"
        },
        {
          "digest": "something else"
        }
      ]
    }
  }
}
```

dockpulp/__init__.py
```diff
@@ -1766,11 +1766,13 @@ def syncRepo(self, env=None, repo=None, config_file=DEFAULT_CONFIG_FILE,
...
         # Need to maintain HIDDEN
-        if repo != HIDDEN:
+        if repo != HIDDEN and (imgs or manifests):
+            pulp_filter = {'unit': {'$or': []}}
             for img in imgs:
-                self.copy(HIDDEN, img, repo)
+                pulp_filter['unit']['$or'].append({'digest': img})
             for manifest in manifests:
-                self.copy(HIDDEN, manifest, repo)
+                pulp_filter['unit']['$or'].append({'manifest_digest': manifest})
```
twaugh: Can we avoid repeating `['unit']['$or']`? e.g. collect the individual unit filters up, then assign at the end?

PR #148, opened 2017-09-27
Fixed breaking change for atomic reactor

breillyr:


breillyr:
@vrutkovs 

Wasn't aware atomic reactor passed a set to crane()

dockpulp/__init__.py
```diff
@@ -1026,7 +1026,7 @@ def copy_filters(self, drepo, source=HIDDEN, filters={}, v1=True, v2=True):
     def crane(self, repos=[], wait=True, skip=False, force_refresh=False):
         """Export pulp configuration to crane for one or more repositories."""
         if not isinstance(repos, list):
-            assert isinstance(repos, str) or isinstance(repos, unicode)
+            assert isinstance(repos, str) or isinstance(repos, unicode) or isinstance(repos, set)
```
lcarva: Doesn't this need to go on the if-statement above?

vrutkovs: Maybe `isinstance(repo, list)` should be `hasattr(repo, '__iter__')`? This way we'll make sure `repos` are iterable, but not `str`?

This looks a bit hackish, but should do the job

lcarva: BTW, `hasattr(repo, '__iter__')` doesn't work on python3.

dockpulp/__init__.py
```diff
@@ -1025,7 +1025,7 @@ def copy_filters(self, drepo, source=HIDDEN, filters={}, v1=True, v2=True):
 
     def crane(self, repos=[], wait=True, skip=False, force_refresh=False):
         """Export pulp configuration to crane for one or more repositories."""
-        if not isinstance(repos, list):
+        if not hasattr(repos, '__iter__'):
```
twaugh: Ick! What about `isinstance(repos, collections.Iterable)`?

vrutkovs: We can't use it, unfortunately, that would return 'True' for strings too

twaugh: Good point. But also, in Python 3:
```
>>> hasattr('', '__iter__')
True
```

vrutkovs: Oh, well, perhaps `isinstance(repos, collections.Iterable) and not ininstance(repos, str)` is the most correct answer?

I think its safe to leave as-is for now, as when py3 comes it would be easy to catch this issue

PR #149, opened 2017-10-05
Changed release options to support rsync distributor

breillyr:


dockpulp/cli.py
```diff
@@ -649,11 +649,16 @@ def do_release(bopts, bargs, parser):
 
     dock-pulp release [options] [repo-id...]
     """
+    parser.add_option('-f', '--force-full', default=False, action='store_true',
```
twaugh: Could include -s and --skip-fast-forward as alternatives for this:
```
parser.add_option('-f', '--force-full', '-s', '--skip-fast-forward', default=False, ...)
```

and remove the explicit option for -s/--skip-fast-forward altogether. Then opts.force_full will be true if any of those four options are provided.

```
$ dock-pulp release --help
Usage: 
    dock-pulp release [options] [repo-id...]

Publish pulp configurations to Crane, making them live. Accepts regex.

Options:
  -h, --help            show this help message and exit
  -f, -s, --force-full, --skip-fast-forward
                        use force_full for release
  -d, -r, --delete, --force-refresh
                        removes extra content on filer that is not in pulp
```

dockpulp/cli.py
```diff
@@ -672,7 +677,8 @@ def do_release(bopts, bargs, parser):
                     rids.extend(results)
             else:
                 rids.append(arg)
-        p.crane(repos=rids, skip=opts.skip, force_refresh=opts.force_refresh)
+        p.crane(repos=rids, delete=(opts.skip or opts.force_full),
+                force_refresh=(opts.force_refresh or opts.delete))
```
twaugh: Should be `delete=...` (the force_refresh keyword argument was removed). With the suggestion about combining the options together this would even be `delete=opts.delete`.

dockpulp/__init__.py
```diff
@@ -1023,7 +1023,7 @@ def copy_filters(self, drepo, source=HIDDEN, filters={}, v1=True, v2=True):
             data=json.dumps(data))
         self.watch(tid)
 
-    def crane(self, repos=[], wait=True, skip=False, force_refresh=False):
+    def crane(self, repos=[], wait=True, force=False, delete=False):
```
twaugh: Removing the old options is fine for atomic-reactor as it only uses 'repos' (positionally) and 'wait' (as a keyword).

Are there any other clients of the dockpulp API that still use the old option names?

dockpulp/cli.py
breillyr: Good point, that's much cleaner

dockpulp/__init__.py
breillyr: I'm not sure, but very true, I'll keep them as they are to avoid breakage.

PR #150, opened 2017-10-09
Added optional config for pulp distributor switchovers

breillyr:
Example of configuration in dockpulp.conf:
```
# Optional section used for Pulp distributor type switchovers                                           
# Define by version number what distributor types should be allowed                                     
# There should be two values per version:                                                               
# First type is the type accepted before the version specified is deployed                              
# Second type is the type accepted after the version specified is deployed                              
[dist_switchover]                                                                                       
2.14 = foo_distributor,bar_distributor
```

### DIFF
--- a/dockpulp/__init__.py
+++ b/dockpulp/__init__.py
@@ -34,6 +34,7 @@ except ImportError:
     gnupg = None
     import subprocess
 from contextlib import closing
+from distutils.version import LooseVersion
 from urlparse import urlparse
 from requests.adapters import HTTPAdapter
 from requests.packages.urllib3.util.retry import Retry
@@ -720,7 +721,8 @@ class Pulp(object):
                               ('retries', "_set_int_attr", "retries"),
                               ('distribution', "_set_bool", "dists"),
                               ('signatures', "_set_independent_attr", "sigs"),
-                              ('sig_exception', "_set_env_attr", "sig_exception"))
+                              ('sig_exception', "_set_env_attr", "sig_exception"),
+                              ('dist_switchover', "_set_independent_attr", "dist_switchover"))
     AUTH_CER_FILE = "pulp.cer"
     AUTH_KEY_FILE = "pulp.key"
 
@@ -752,20 +754,16 @@ class Pulp(object):
         if not os.path.exists(config_distributions):
                 log.error('could not load distributions json: %s' % config_distributions)
         self.distributionconf = json.load(open(config_distributions, 'r'))
-        try:
-            self.timeout
-        except AttributeError:
+        if not hasattr(self, 'timeout'):
             self.timeout = 180
         if self.timeout is None:
             self.timeout = 180
-        try:
-            self.dists
-        except AttributeError:
+        if not hasattr(self, 'dists'):
             self.dists = False
-        try:
-            self.sig_exception
-        except AttributeError:
+        if not hasattr(self, 'sig_exception'):
             self.sig_exception = None
+        if not hasattr(self, 'dist_switchover'):
+            self.dist_switchover = {}
 
     def _set_bool(self, attrs):
         for key, boolean in attrs:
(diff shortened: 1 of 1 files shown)


dockpulp/__init__.py
```diff
@@ -766,6 +768,10 @@ def __init__(self, env='qa', config_file=DEFAULT_CONFIG_FILE,
             self.sig_exception
         except AttributeError:
             self.sig_exception = None
+        try:
+            self.dist_switchover
+        except AttributeError:
```
twaugh: or `if not hasattr(self, 'dist_switchover')`

dockpulp/__init__.py
```diff
@@ -1162,9 +1168,24 @@ def createRepo(self, repo_id, url, registry_id=None, desc=None, title=None, prot
         if self.distributors == "":
             distributors = False
         if distributors:
+            type_id = None
+            # dist_switchover is an optional section in dockpulp.conf
+            # Used for changes in Pulp distributor types
+            for key in self.dist_switchover:
+                before_id = self.dist_switchover[key].split(",")[0]
+                after_id = self.dist_switchover[key].split(",")[1]
+                if LooseVersion(self.getPulpVersion()) >= LooseVersion(key):
```
twaugh: This makes a Pulp version request every time we go round the loop. Call the method first, put the result in a local variable, then loop.

dockpulp/__init__.py
```diff
@@ -1162,9 +1168,24 @@ def createRepo(self, repo_id, url, registry_id=None, desc=None, title=None, prot
...
+                    type_id = after_id
```
twaugh: Do we want to break out of the loop here?

dockpulp/__init__.py
```diff
@@ -1162,9 +1168,24 @@ def createRepo(self, repo_id, url, registry_id=None, desc=None, title=None, prot
...
             stuff['distributors'] = []
             distributorkeys = self.distributors.strip().split(",")
             for key in distributorkeys:
+                if type_id:
+                    dtype = self.distributorconf[key]['distributor_type_id']
+                    # exclude default distributors
+                    if dtype == before_id:
```
twaugh: We need to make sure `before_id` definitely gets set. Right now if `self.dist_switchover` is an empty iterable it won't be.

dockpulp/__init__.py
```diff
@@ -1162,9 +1160,29 @@ def createRepo(self, repo_id, url, registry_id=None, desc=None, title=None, prot
...
             distributors = False
         if distributors:
+            type_id = None
+            before_id = None
+            after_id = None
+            # dist_switchover is an optional section in dockpulp.conf
+            # Used for changes in Pulp distributor types
+            if self.dist_switchover:
+                pulpversion = self.getPulpVersion()
+            for key in self.dist_switchover:
```
twaugh: Since we immediately look up the key, we could use .items(): `for key, value in self.dist_switchover.items():`

dockpulp/__init__.py
```diff
@@ -1162,9 +1160,29 @@ def createRepo(self, repo_id, url, registry_id=None, desc=None, title=None, prot
...
+                before_id = self.dist_switchover[key].split(",")[0]
+                after_id = self.dist_switchover[key].split(",")[1]
```
twaugh: Should this be `.split(",", 1)[1]`? Alternatively, should we validate the value, that it only has a single comma?

dockpulp/__init__.py
```diff
@@ -1162,9 +1160,29 @@ def createRepo(self, repo_id, url, registry_id=None, desc=None, title=None, prot
...
+                if LooseVersion(pulpversion) >= LooseVersion(key):
+                    type_id = after_id
```
twaugh: Since we only use `after_id` and `before_id` if we get here, should we move the string processing to here?

dockpulp/__init__.py
twaugh: This is fine now.

dockpulp/__init__.py
```diff
@@ -1162,9 +1160,29 @@ def createRepo(self, repo_id, url, registry_id=None, desc=None, title=None, prot
...
+                    break
             stuff['distributors'] = []
             distributorkeys = self.distributors.strip().split(",")
             for key in distributorkeys:
+                if type_id:
+                    dtype = self.distributorconf[key]['distributor_type_id']
+                    # exclude default distributors
+                    if dtype == before_id:
+                        log.debug('Using new distributor type %s for distributor %s' %
+                                  (type_id, key))
```
twaugh: Better to pass logging values as separate parameters: `log.debug('....', type_id, key)`

dockpulp/__init__.py
```diff
@@ -1162,9 +1160,31 @@ def createRepo(self, repo_id, url, registry_id=None, desc=None, title=None, prot
...
+            # Used for changes in Pulp distributor types
+            if self.dist_switchover:
+                pulpversion = self.getPulpVersion()
+            for key, value in self.dist_switchover.items():
+                if LooseVersion(pulpversion) >= LooseVersion(key):
+                    if value.count(',') != 1:
+                        raise errors.DockPulpConfigError(
+                            'dist_switchover must be a comma separated list of two values')
+                    before_id = value.split(",")[0]
+                    after_id = value.split(",")[1]
```
twaugh: Nit: just spotted this could be `before_id, after_id = value.split(",")` now that we've validated it :)

PR #151, opened 2017-10-12
Log the actual status code when >= 500

twaugh:
Signed-off-by: Tim Waugh <twaugh@redhat.com>


dockpulp/__init__.py
```diff
@@ -148,8 +148,6 @@ def __call__(self, meth, api, **kwargs):
             log.warning('No content in Pulp response')
         if answer.status_code == 403:
             raise errors.DockPulpLoginError('Received 403: Forbidden')
-        elif answer.status_code >= 500:
```
lcarva: I don't get it. There's no default `else` statement here. Wouldn't this change just ignore 5XX errors?

lcarva: Ah doh! 500 >= 400

PR #154, opened 2017-10-23
Restrict repo id based on crane requirements

breillyr:


dockpulp/__init__.py
```diff
@@ -1076,6 +1076,14 @@ def createRepo(self, repo_id, url, registry_id=None, desc=None, title=None, prot
 
         id and description are required
         """
+        # restrict repoid names based on Crane requirements
+        restricted = ['tags', 'manifests', 'blobs']
+        for r in restricted:
+            if r in repo_id:
```
twaugh: My understanding is that it's only things like "tags/foo" and "foo/tags" that would fail, not things like "hashtags/search" or "ghostly/manifestation", i.e. only pathname components need to be checked.

Is this check too restrictive? If so, maybe it could be done like this:
```
repo_id_components = repo_id.split('/')
restricted = ...
for r in restricted:
    if r in repo_id_components:
        ...
```

PR #155, opened 2017-10-24
Added check to dockpulp login to error on incorrect input

breillyr:


dockpulp/cli.py
```diff
@@ -615,6 +615,9 @@ def do_login(bopts, bargs, parser):
     parser.add_option('-p', '--password', help='specify an account password')
     parser.add_option('-u', '--username', default='admin', help='pick username')
     opts, args = parser.parse_args(bargs)
+    if len(args) > 0:
+        log.error("No arguments should be passed to login. Use the -p and -u options.")
+        sys.exit(1)
```
twaugh: Does this want a test_do_login() in tests/test_cli.py?

tests/test_cli.py
```diff
@@ -293,3 +297,32 @@ def test_do_empty(self, bargs):
...
+                .new_instances(p)
+                .with_args(Pulp, env=bopts.server, config_file=bopts.config_file))
+            bargs.append('--password=foobar')
+            (flexmock(os.path)
+                .should_receive('expanduser')
+                .once()
+                .and_return("dir"))
+            (flexmock(os.path)
+                .should_receive('exists')
+                .at_least.once()
```
twaugh: Looks like this should be `.at_least().once()` -- or, alternatively, from a behaviour testing point of view we don't care about whether os.path.exists() gets called and only mock it if it is, i.e. flexmock(os.path).should_receive('exists').and_return(True).

For example, an implementation might simply os.mkdir() and catch any exception it throws when the directory already exists, and we'd still want that to pass unit testing.

lcarva:
@breillyr, looks like some comments haven't been addressed?

PR #156, opened 2017-11-03
Added switchover capability for sigstore distributors

breillyr:


breillyr:
Not sure of a good way to test dock-pulp-bootstrap as it's not part of the module, but I added a test for the associate change.

PR #159, opened 2017-11-15
Added switchover func for release order

breillyr:


twaugh:
I don't understand what this does. :-)

Should there be an update to conf/dockpulp.conf demoing what these look like?

Can the unit tests be extended to cover these new lines?

breillyr:
This is to add functionality for automatic switching of release order based on pulp version, in case the old release orders are deprecated.

tests/test_pulp_instance.py
```diff
@@ -185,6 +185,78 @@ def restricted_pulp(tmpdir):
     return restricted_pulp
 
 
+@pytest.fixture
```
twaugh: A fixture is a function which py.test runs in order to provide a parameter to a test function (or method) -- the fixture's return value provides the parameter value.

Since no test function takes a parameter named `switched_pulp`, py.test never calls this function.

What you want here is a test function (or method), i.e. don't decorate it with `@pytest.fixture`, and name it anything starting with `test_`.

Incidentally, the same goes for `core_pulp` -- nothing calls that, for the same reason.

breillyr: Hm, I could have sworn core_pulp was used in other tests, I created it to test cases where optional config doesn't exist. However it looks like nothing does actually use it, though I still think it's valuable to leave there for the future.

tests/test_pulp_instance.py
```diff
@@ -219,6 +219,76 @@ def hexdigest(self):
 # tests
 class TestPulp(object):
     # Tests of methods from Pulp class.
+    def test_switched_pulp(tmpdir):
```
twaugh: Should be `self` instead of `tmpdir`. Once that's fixed this looks good to go.

PR #161, opened 2017-12-19
Added support for listing and confirming manifest lists

breillyr:


dockpulp/__init__.py
```diff
@@ -451,53 +456,70 @@ def _test_repo(self, dockerid, redirect, pulp_imgs, protected=False, silent=Fals
...
+        s2_blobs_to_test = set()
         try:
             for manifest in pulp_manifests:
-                answer = requests.get(url + '/' + manifest, verify=False,
-                                      cert=(self.cert, self.key))
-                log.debug('  crane content: %s', answer.content)
-                log.debug('  status code: %s', answer.status_code)
+                schema_ver = repo['manifests'][manifest]['schema_version']
+                if schema_ver == 1:
+                    answer = requests.get(url + '/' + manifest, verify=False,
```
twaugh: Why not `'/1/'`?

dockpulp/__init__.py
```diff
@@ -451,53 +456,70 @@ def _test_repo(self, dockerid, redirect, pulp_imgs, protected=False, silent=Fals
...
-                    log.error('  Incorrect name (%s) in manifest: %s',
-                              manifest_name, manifest)
-                    result['error'] = True
-                    result['crane_manifests_incorrectly_named'].append(manifest)
+                if schema_ver == 1:
+                    for fs_layer in manifest_json['fsLayers']:
+                        blobs_to_test.add(fs_layer['blobSum'])
+                else:
+                    for fs_layer in manifest_json['layers']:
+                        s2_blobs_to_test.add(fs_layer['digest'])
```
twaugh: How about the config blob?

dockpulp/__init__.py
```diff
@@ -563,37 +629,70 @@ def _test_repoV2(self, dockerid, redirect, pulp_manifests, pulp_blobs, pulp_tags
         log.info('  Expected and available blobs reconciled correctly, testing tags')
 
         url = self.p.registry + '/v2/' + dockerid + '/tags/list'
+        s2url = self.p.registry + '/pulp/docker/v2/' + repoid + '/tags/list'
```
twaugh: Is that meant to be a Docker Registry API crane URL (as the debug message shows below), or a Pulp data-store URL, which is what it looks like?

dockpulp/__init__.py
```diff
@@ -451,53 +456,70 @@ def _test_repo(self, dockerid, redirect, pulp_imgs, protected=False, silent=Fals
...
+    def _test_repoV2(self, repo, dockerid, repoid, redirect, pulp_manifests, pulp_manifest_lists,
+                     pulp_blobs, pulp_tags, protected=False, silent=False):
         """Confirm we can reach crane and get data back from it."""
         result = {}
         result['error'] = False
         if not pulp_manifests:
             log.info('  No v2 content to test')
             return result
         url = self.p.registry + '/v2/' + dockerid + '/manifests'
+        s2url = self.p.registry + '/pulp/docker/v2/' + repoid + '/manifests'
```
twaugh: Huh? That doesn't make much sense to me. What is this meant to be?

twaugh:
I think this could definitely do with some clarification about exactly what it is intending to check. The debug/warning messages don't seem to match what's actually being tested.

dockpulp/__init__.py
breillyr: Schema 2 uses a different url to get at information in crane, including changing the reponame from docker id (test/repo) to repo id (redhat-test-repo)

twaugh: No, that's not a difference between schema 1 and schema 2. That's a difference between the Docker Registry API and the Pulp data-store.

dockpulp/__init__.py
```diff
@@ -202,11 +202,15 @@ def confirm(self, repos, v1=True, v2=True, silent=True, check_layers=False):
                 continue
             imgs = repo['images'].keys()
             manifests = repo['manifests'].keys()
+            manifest_lists = repo['manifest_lists'].keys()
```
twaugh: Is this ever used?

dockpulp/__init__.py
```diff
@@ -476,28 +481,39 @@ def _test_repoV2(self, dockerid, redirect, pulp_manifests, pulp_blobs, pulp_tags
         result['crane_manifests_incorrectly_named'] = []
         blobs_to_test = set()
         try:
+            s = requests.Session()
             for manifest in pulp_manifests:
-                answer = requests.get(url + '/' + manifest, verify=False,
-                                      cert=(self.cert, self.key))
+                schema_ver = repo['manifests'][manifest]['schema_version']
+                if schema_ver == 1:
+                    s.headers['Accept'] = '*/*'
```
twaugh: Or just don't set the Accept header in this case.

dockpulp/__init__.py
```diff
@@ -769,7 +824,7 @@ def __init__(self, env='qa', config_file=DEFAULT_CONFIG_FILE,
                 raise errors.DockPulpConfigError('Only one switchover version can be defined')
             for val in self.switch_ver.values():
                 if LooseVersion(self.getPulpVersion()) >= LooseVersion(val):
-                    if hasattr(self, 'switch_release'):
+                    if hasattr(self, 'switch_release') and self.switch_release is not None:
```
twaugh: This seems unrelated?

dockpulp/__init__.py
```diff
@@ -476,28 +481,39 @@ def _test_repoV2(self, dockerid, redirect, pulp_manifests, pulp_blobs, pulp_tags
...
+                else:
+                    s.headers['Accept'] = 'application/vnd.docker.distribution.manifest.v2+json'
+                answer = s.get(url + '/' + manifest, verify=False,
+                               cert=(self.cert, self.key))
                 log.debug('  crane content: %s', answer.content)
                 log.debug('  status code: %s', answer.status_code)
                 if not answer.ok:
                     continue
-
                 c_manifests.add(manifest)
```
twaugh: Before counting this as found, we should verify that it's the type we asked for i.e. that `answer.json()['mediaType']` is the same as what our Accept header asked for (this test works even for Pulp < 2.15).

dockpulp/__init__.py
```diff
@@ -523,6 +539,45 @@ def _test_repoV2(self, dockerid, redirect, pulp_manifests, pulp_blobs, pulp_tags
...
+            s.headers['Accept'] = 'application/vnd.docker.distribution.manifest.list.v2+json'
+            for manifest_list in pulp_manifest_lists:
+                answer = s.get(url + '/' + manifest_list, verify=False,
+                               cert=(self.cert, self.key))
+                log.debug('  crane content: %s', answer.content)
+                log.debug('  status code: %s', answer.status_code)
+                if not answer.ok:
+                    continue
+
+                c_manifest_lists.add(manifest_list)
```
twaugh: Before counting this as found, we should verify that it's the type we asked for i.e. that `answer.json()['mediaType']` is the same as what our Accept header asked for.

dockpulp/__init__.py
```diff
@@ -202,11 +202,15 @@ def confirm(self, repos, v1=True, v2=True, silent=True, check_layers=False):
...
             tags = []
-            for manifest in manifests:
-                blobs.extend(repo['manifests'][manifest]['layers'])
-                tags.append(repo['manifests'][manifest]['tag'])
+            for manifest_list in repo['manifest_lists'].values():
+                tags.extend(manifest_list['tags'])
+            for manifest in repo['manifests'].values():
+                blobs.extend(manifest['layers'])
+                if manifest['tag']:
+                    tags.append(manifest['tag'])
```
twaugh: For each tag we want to test, we should really test that it can reach a manifest of the expected type. Currently I think we'll just test that all of them can reach a v2 schema 1 manifest -- including the identically-named tags intended for other content.

I'd suggest making `tags` a dict of tag name to an iterable of media types, and having the confirm function iterate over the media types and confirm the tag returns the correct object when the Accept header is set to each.

dockpulp/__init__.py
lcarva: It's being set on the session which is reused between loop iterations.

dockpulp/__init__.py
breillyr: Yes, bad patch form on my part, this line was a small bugfix regarding switching the release order.

dockpulp/__init__.py
twaugh: It's just not clear to me how a registry it supposed to respond to */*.

dockpulp/__init__.py
breillyr: Yes, it's passed to _test_repoV2

twaugh: Would be easier to read if it was closer to that method call.

PR #162, opened 2017-12-20
Made manifest list output an option

breillyr:


twaugh:
I'm not sure I understand why this is something that should be optional.

twaugh:
Also, if you really want to make it optional, you should remove the list type from the query if not required.

PR #163, opened 2018-01-02
Manifest List Support Improvements

breillyr:
Mlists now group correctly in output
Added support for mediatype checking in confirm
Fixed bug with silent confirms not showing correct number of failing repos

### DIFF
--- a/dockpulp/__init__.py
+++ b/dockpulp/__init__.py
@@ -479,21 +479,31 @@ class Crane(object):
                 return result
 
         result['crane_manifests_incorrectly_named'] = []
+        result['incorrect_mediatype'] = []
         blobs_to_test = set()
         try:
             s = requests.Session()
+            # mediatype for schema 2 manifests
+            mediatype = 'application/vnd.docker.distribution.manifest.v2+json'
             for manifest in pulp_manifests:
                 schema_ver = repo['manifests'][manifest]['schema_version']
                 if schema_ver == 1:
                     s.headers['Accept'] = '*/*'
                 else:
-                    s.headers['Accept'] = 'application/vnd.docker.distribution.manifest.v2+json'
+                    s.headers['Accept'] = mediatype
                 answer = s.get(url + '/' + manifest, verify=False,
                                cert=(self.cert, self.key))
                 log.debug('  crane content: %s', answer.content)
                 log.debug('  status code: %s', answer.status_code)
                 if not answer.ok:
                     continue
+
+                if schema_ver == 2 and answer.json()['mediaType'] != mediatype:
+                    log.error('  Incorrect mediatype for schema 2 manifest: %s %s',
+                              manifest, answer.json()['mediaType'])
+                    result['error'] = True
+                    result['incorrect_mediatype'].append(manifest)
+
                 c_manifests.add(manifest)
 
                 manifest_json = answer.json()
@@ -541,9 +551,10 @@ class Crane(object):
 
         log.info('  Testing Pulp and Crane manifest lists')
         c_manifest_lists = set()
+        mediatype = 'application/vnd.docker.distribution.manifest.list.v2+json'
         try:
             s = requests.Session()
-            s.headers['Accept'] = 'application/vnd.docker.distribution.manifest.list.v2+json'
+            s.headers['Accept'] = mediatype
             for manifest_list in pulp_manifest_lists:
                 answer = s.get(url + '/' + manifest_list, verify=False,
                                cert=(self.cert, self.key))
@@ -551,6 +562,12 @@ class Crane(object):
                 log.debug('  status code: %s', answer.status_code)
                 if not answer.ok:
                     continue
+
+                if answer.json()['mediaType'] != mediatype:
+                    log.error('  Incorrect mediatype manifest list: %s %s',
+                              manifest_list, answer.json()['mediaType'])
+                    result['error'] = True
+                    result['incorrect_mediatype'].append(manifest)
 
                 c_manifest_lists.add(manifest_list)
 
--- a/dockpulp/__init__.py
+++ b/dockpulp/__init__.py
@@ -197,7 +197,7 @@ class Crane(object):
                     repoids[repo['id']].update(response)
                     if response['error']:
                         errorids[repo['id']] = True
-                elif response['error']:
+                if response['error']:
                     errors += 1
                 continue
             imgs = repo['images'].keys()
@@ -220,7 +220,7 @@ class Crane(object):
                     repoids[repo['id']].update(response)
                     if response['error']:
                         errorids[repo['id']] = True
-                elif response['error']:
+                if response['error']:
                     errors += 1
             if v2 == auto:
                 log.debug('  Checking whether v2 is supported by crane')
@@ -238,7 +238,7 @@ class Crane(object):
                     repoids[repo['id']].update(response)
                     if errorids[repo['id']]:
                         repoids[repo['id']]['error'] = True
-                elif response['error']:
+                if response['error']:
                     errors += 1
 
             if check_layers:
@@ -250,7 +250,7 @@ class Crane(object):
                         repoids[repo['id']].update(response)
                         if errorids[repo['id']]:
                             repoids[repo['id']]['error'] = True
-                    elif response['error']:
+                    if response['error']:
                         errors += 1
 
                 if v2:
@@ -259,7 +259,7 @@ class Crane(object):
                         repoids[repo['id']].update(response)
                         if errorids[repo['id']]:
                             repoids[repo['id']]['error'] = True
-                    elif response['error']:
+                    if response['error']:
                         errors += 1
 
         repoids['numerrors'] = errors
--- a/dockpulp/cli.py
+++ b/dockpulp/cli.py
@@ -451,6 +451,25 @@ def do_imageids(bopts, bargs, parser):
     log.info(result)
 
 
+def _list_manifest_helper(output, manifest, schema):
+    tagoutput = None
+    tag = output['manifests'][manifest]['tag']
+    is_active = output['manifests'][manifest]['active']
+    if tag is None:
+        log.info('  Manifest: %s', manifest)
+    else:
+        log.info('  Manifest: %s  Tag: %s%s', manifest, tag, is_active)
+        if is_active:
+            tagoutput = tag
+        config = output['manifests'][manifest]['config']
+        if config:
+            log.info('    Config Layer: %s', config)
+        sv = output['manifests'][manifest]['sv']
+        if schema and sv:
+            log.info('    Schema Version: %s', sv)
+    return tagoutput
+
+
 @make_parser
 def do_list(bopts, bargs, parser):
     """List one or more repositories. Accepts regex.
@@ -548,12 +567,20 @@ def do_list(bopts, bargs, parser):
                 manifest_lists.sort()
                 tags = repo['tags']
                 output = {}
+                seenmanifests = {}
+                seenlayers = {}
+                manifest_layers = {}
 
                 for manifest in manifests:
+                    seenmanifests[manifest] = False
+
                     layer = tuple(repo['manifests'][manifest]['layers'])
+                    seenlayers[layer] = False
 
                     output.setdefault(layer, {'manifests': {}})
                     output[layer]['manifests'].setdefault(manifest, {})
+
+                    manifest_layers[manifest] = layer
 
                     tag = repo['manifests'][manifest]['tag']
                     output[layer]['manifests'][manifest]['tag'] = tag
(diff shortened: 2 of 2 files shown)


dockpulp/cli.py
```diff
@@ -577,13 +577,12 @@ def do_list(bopts, bargs, parser):
                     layer = tuple(repo['manifests'][manifest]['layers'])
                     seenlayers[layer] = False
 
-                    output.setdefault(layer, {'manifests': {}})
-                    output[layer]['manifests'].setdefault(manifest, {})
-
+                    output.setdefault(layer, {})
+                    output[layer].setdefault(manifest, {})
```
twaugh: I'm really unclear on what 'output' actually is, and how it relates to output from the CLI. Is it just tracking what we've already reported?

Part of the problem is that the return from listRepos is different from what the Pulp response looks like, but the schema returned by listRepos is not described anywhere, so it's hard to review changes to the code.

twaugh:
There are no unit tests.

dockpulp/cli.py
```diff
@@ -451,6 +451,25 @@ def do_imageids(bopts, bargs, parser):
...
 
 
+def _list_manifest_helper(output, manifest, schema):
+    tagoutput = None
+    tag = output[manifest]['tag']
+    is_active = output[manifest]['active']
+    if tag is None:
+        log.info('  Manifest: %s', manifest)
+    else:
+        log.info('  Manifest: %s  Tag: %s%s', manifest, tag, is_active)
```
lcarva: Won't this print something like:
```
Manifest: sha256:xxxxx...xx Tag: latestTrue
```

A better output might be:
```
Manifest: sha256:xxxxx...xx Tag: latest (active)
```
Similarly to how other tags are listed.

dockpulp/cli.py
```diff
@@ -568,75 +594,77 @@ def do_list(bopts, bargs, parser):
...
-                                if manifest in repo['manifest_lists'][manifest_list]['mdigests']:
-                                    output[layer]['manifest_list']['digest'] = manifest_list
-                                    output[layer]['manifest_list']['exist'] = True
-                                mltags = repo['manifest_lists'][manifest_list]['tags']
-                                output[layer]['manifest_list']['tags'] = ', '.join(mltags)
+                        output[layer][manifest]['id'] = v1id
+                        output[layer][manifest]['parent'] = v1parent
+                        output[layer][manifest]['labels'] = v1labels
+
+                if opts.lists:
```
lcarva: I'm having trouble reviewing changes in this method because I'm not sure what the data structure looks like.
We should at least document this somewhere, especially because this data structure is exposed via the `silent` option.

dockpulp/cli.py
breillyr: Output is a restructuring of the information received from listRepos to allow for readable output - grouping manifests by their shared blobs etc.

dockpulp/cli.py
breillyr: is_active is a string, either '' or ' (active)', so the output would be correct.

dockpulp/__init__.py
```diff
@@ -479,21 +479,31 @@ def _test_repoV2(self, repo, dockerid, repoid, redirect, pulp_manifests, pulp_ma
...
-                    s.headers['Accept'] = 'application/vnd.docker.distribution.manifest.v2+json'
+                    s.headers['Accept'] = mediatype
                 answer = s.get(url + '/' + manifest, verify=False,
                                cert=(self.cert, self.key))
                 log.debug('  crane content: %s', answer.content)
                 log.debug('  status code: %s', answer.status_code)
                 if not answer.ok:
                     continue
+
+                if schema_ver == 2 and answer.json()['mediaType'] != mediatype:
```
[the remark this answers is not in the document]
breillyr: 'mediaType' does not seem to exist in the response for schema 1 images.

dockpulp/cli.py
```diff
@@ -548,15 +567,22 @@ def do_list(bopts, bargs, parser):
...
                 output = {}
+                seenmanifests = {}
+                seenlayers = {}
+                manifest_layers = {}
 
                 for manifest in manifests:
+                    seenmanifests[manifest] = False
+
                     layer = tuple(repo['manifests'][manifest]['layers'])
+                    seenlayers[layer] = False
```
twaugh: Oh! Is `layer` actually a tuple of layers? That wasn't obvious to me at first read. Can we rename that variable if so ("layers")?

dockpulp/cli.py
```diff
@@ -568,75 +594,77 @@ def do_list(bopts, bargs, parser):
...
-                        if output[layer]['manifests'][manifest]['sv'] == 2:
-                            for manifest_list in manifest_lists:
-                                if manifest in repo['manifest_lists'][manifest_list]['mdigests']:
-                                    output[layer]['manifest_list']['digest'] = manifest_list
-                                    output[layer]['manifest_list']['exist'] = True
-                                mltags = repo['manifest_lists'][manifest_list]['tags']
-                                output[layer]['manifest_list']['tags'] = ', '.join(mltags)
+                        output[layer][manifest]['id'] = v1id
+                        output[layer][manifest]['parent'] = v1parent
+                        output[layer][manifest]['labels'] = v1labels
```
twaugh: This is a lot of lookups, and a lot to read. Can we simplify it please? Maybe build up the dictionary first, then finally `output[layer][manifest] = metadata`?

Also, is there anything we are *not* copying from `repo['manifests'][manifest]`?

dockpulp/cli.py
```diff
@@ -548,15 +567,22 @@ def do_list(bopts, bargs, parser):
                 manifest_lists.sort()
                 tags = repo['tags']
                 output = {}
+                seenmanifests = {}
+                seenlayers = {}
+                manifest_layers = {}
 
                 for manifest in manifests:
+                    seenmanifests[manifest] = False
```
twaugh: If we're just using this to work out if we've processed a particular manifest digest before, why not make sure `manifests` doesn't contain duplicates? `for manifest in set(manifests):`

Is there any particular reason to think it *might* contain duplicates?

dockpulp/cli.py
twaugh: If we're just using this to determine whether we've processed a particular tuple of blob digests before, why not initialise `seenlayers` as a `set()` and `seenlayers.add()` each tuple we've processed? Then the "have I seen this before?" test is `if layer in seenlayers` (but NB, let's change the variable name 'layer' to 'layers').

dockpulp/cli.py
```diff
@@ -568,75 +594,77 @@ def do_list(bopts, bargs, parser):
...
-                                mltags = repo['manifest_lists'][manifest_list]['tags']
-                                output[layer]['manifest_list']['tags'] = ', '.join(mltags)
+                        output[layer][manifest]['id'] = v1id
+                        output[layer][manifest]['parent'] = v1parent
+                        output[layer][manifest]['labels'] = v1labels
+
+                if opts.lists:
+                    mloutput = {}
+                    for manifest_list in manifest_lists:
+                        mdigest = tuple(repo['manifest_lists'][manifest_list]['mdigests'])
```
twaugh: This is another instance of singular/plural confusion. Is it a digest, or some digests? Let's be consistent with the naming as it really helps readability.

dockpulp/cli.py
```diff
@@ -568,75 +594,77 @@ def do_list(bopts, bargs, parser):
...
+                        log.info('')
+                        mltags = mloutput[mdigest]['tags']
+                        log.info('  Manifest List: %s', mloutput[mdigest]['mldigest'])
+                        if tags:
+                            log.info('    Tags: %s', mltags)
+                        for manifest in mdigest:
+                            seenmanifests[manifest] = True
+                            layers = manifest_layers[manifest]
+                            seenlayers[layers] = True
+                            _list_manifest_helper(output[layers], manifest, opts.schema)
```
twaugh: Helper? What does it actually do? Can we come up with a better name?

dockpulp/cli.py
```diff
@@ -568,75 +594,77 @@ def do_list(bopts, bargs, parser):
...
-                                    output[layer]['manifest_list']['digest'] = manifest_list
-                                    output[layer]['manifest_list']['exist'] = True
-                                mltags = repo['manifest_lists'][manifest_list]['tags']
-                                output[layer]['manifest_list']['tags'] = ', '.join(mltags)
+                        output[layer][manifest]['id'] = v1id
+                        output[layer][manifest]['parent'] = v1parent
+                        output[layer][manifest]['labels'] = v1labels
+
+                if opts.lists:
+                    mloutput = {}
```
twaugh: This seems to be a manifest-list version of 'output', so the same concerns about naming and description apply.

dockpulp/cli.py
```diff
@@ -568,75 +594,77 @@ def do_list(bopts, bargs, parser):
...
+                    for manifest_list in manifest_lists:
+                        mdigest = tuple(repo['manifest_lists'][manifest_list]['mdigests'])
+                        mloutput.setdefault(mdigest, {})
+                        mloutput[mdigest]['mldigest'] = manifest_list
+                        mltags = repo['manifest_lists'][manifest_list]['tags']
+                        mloutput[mdigest]['tags'] = ', '.join(mltags)
```
twaugh: I have no idea what this loop is for. `mloutput` seems to be a dictionary mapping image manifest digests tuples to manifest list digests. I don't understand why we need/want to do that.

dockpulp/cli.py
```diff
@@ -568,75 +594,77 @@ def do_list(bopts, bargs, parser):
...
+
+                    mdigests = mloutput.keys()
+                    for mdigest in mdigests:
```
twaugh: We use the value as well in this loop, so we can call items() instead of keys(). Also, the keys are tuples of digests not single digests, so let's use a plural name:
`for mdigests, metadata in mloutput.items():`

dockpulp/cli.py
```diff
@@ -568,75 +594,77 @@ def do_list(bopts, bargs, parser):
...
-                                output[layer]['manifest_list']['tags'] = ', '.join(mltags)
+                        output[layer][manifest]['id'] = v1id
+                        output[layer][manifest]['parent'] = v1parent
+                        output[layer][manifest]['labels'] = v1labels
+
+                if opts.lists:
+                    mloutput = {}
+                    for manifest_list in manifest_lists:
+                        mdigest = tuple(repo['manifest_lists'][manifest_list]['mdigests'])
+                        mloutput.setdefault(mdigest, {})
```
twaugh: We always set the same keys, so instead we can just assign it.
```
mloutput[mdigest] = {
    'mldigest': ...,
    ...
}
```

dockpulp/cli.py
breillyr: The purpose of this loop is to do map manifest list digests to manifest digest tuples. The reason why we want this is because tuples can be shared between manifest lists and we want to capture that in our output.

For example, let's say we have manifest list A and manifest list B, and each of them references manifests (X, Y, Z). If I just took the output from listRepo, which maps manifest digests to manifest lists, the resulting output would look like:
Manifest List A
    (X, Y, Z)
Manifest List B
    (X, Y, Z)

Whereas we want something more like this:
Manifest List A
Manifest List B
    (X, Y, Z)

There may be other ways to have output like the above - I'm open to suggestions on this, this solution is admittedly very messy.

dockpulp/cli.py
breillyr: Output's purpose is to restructure the information we got from listRepo and output it in a human readable manner.
The biggest thing here is mapping the layer tuple -> manifest digest. You can have multiple manifests with the same layers, and the desired output is to group those together. Say you have Manifest A and Manifest B and layers (X, Y, Z).

If I printed it out while using the dictionary structure we got back from listRepos I would get something like:
Manifest A
    (X, Y, Z)
Manifest B
    (X, Y, Z)

Whereas what we want is
Manifest A
Manifest B
    (X, Y, Z)

The rest of the structure is identical to the dict listRepos returns otherwise, but grouping manifests by tuple allows for iteration over *tuple*, this guaranteeing we'll be able to group up the manifests accordingly in output.

dockpulp/cli.py
breillyr: It's not duplicates we're worried about here, it's because some manifests can be referenced in multiple places.
Namely, the manifest itself, and then possible references *to* that manifest in manifest lists. 
seenmanifests (as well as seenlayers) gets set to True if it is seen while printing out manifest list information.
This means that any manifest or layer tuple that is seen while printing out manifest lists gets skipped over while
iterating through the manifests lists that are *not* grouped up under a manifest list.

These variables are ignored when not using the --list option.

dockpulp/cli.py
breillyr: it prints out manifest information, including the digest, tags, config layer and schema version

dockpulp/cli.py
breillyr: seenlayers is used for the same reason as seenmanifest - namely that a tuple of layers may be referenced as part of a manifest that is referenced in a manifest list - in these cases, we only want to print out the manifest and associated tuple *once*, grouped with its manifest list. Without this, the manifest list / manifest / layers output would print out, and then that same manifest / layers output would print below that, duplicating our output.

Identical tuple layers at this step should not be a problem because of the setdefault immediately aftewards - the expected layer tuple duplicates are handled automatically.

dockpulp/__init__.py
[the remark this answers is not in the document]
breillyr: Just checked it - that "Content-Type" field only seems to appear in response headers for Schema 2 manifests.
Schema 1 manifests just have Content-Length, Accept-Ranges, Keep-Alive, Server, Last-Modified, Connection, Etag, Docker-Distribution-API-Version and Date in their headers. Schema 2 has all of those, with Content-Type being the only addition.

breillyr:
Looks like the comments from before the last 4 commits auto-hid, just fyi I answered some questions there, though hopefully my intent is more obvious in the code/comments now.

twaugh:
> For example, let's say we have manifest list A and manifest list B, and each of them references manifests (X, Y, Z). If I just took the output from listRepo, which maps manifest digests to manifest lists, the resulting output would look like:
Manifest List A
(X, Y, Z)
Manifest List B
(X, Y, Z)
> 
> Whereas we want something more like this:
Manifest List A
Manifest List B
(X, Y, Z)
> 
> There may be other ways to have output like the above - I'm open to suggestions on this, this solution is admittedly very messy.

I think that situation is unlikely, and here's why. A manifest list is very little more than just the list of the manifest digests. The only things you can change in manifest list A to make a different manifest list B are

- the platform specifiers, i.e. which platform does the image for this manifest run on?
- the order in which the manifest digests are listed
- add or remove a manifest digest

Changing the platform specifier might feasibly happen, although it's unlikely. Instead of saying something lik2 "this runs on amd64" it would be something like saying "this runs on amd64, but only on CPUs with built-in Meltdown protection" or whatever, i.e. some sort of additional 'feature'.

The order might be changed if two platform specifiers for the same CPU architecture wanted different 'features', and we wanted to change the priority of the features. This is quite a contrived example.

Manifest list B having a subset of the image manifests from manifest list A (say) *is* quite a likely scenario, for instance when an image is temporarily built without support for some architecture, but it is later added (e.g. after testing).

But in all of these cases I feel like if we were to only list the contained manifests once then we would be missing some information. It is quite a different thing than having two different image manifests having the same filesystem layer blobs, which must always appear in the same order and which carry no additional about the blobs in the manifest.

breillyr:
Sorry that was a mistake on my part regarding the manifest lists - it won't print out all shared manifests as
Manifest List A
Manifest List B
(manifests)

It will instead do
Manifest List A
X Y Z

Manifest List B
X Y Z

The purpose for grouping up manifest lists by a tuple of manifests was actually to stop output like this from occurring:
Manifest List A
X
Manifest List A
Y
Manifest List A
Z

Even if two manifest lists share the same manifests tuple they will be printed out separately.

dockpulp/cli.py
```diff
@@ -451,6 +451,151 @@ def do_imageids(bopts, bargs, parser):
...
+            mloutput[mdigests] = {
+                'mldigest': manifest_list,
+                'tags': ', '.join(mltags)
+            }
+
+            for mdigests, listinfo in mloutput.items():
+                log.info('')
+                mltags = listinfo['tags']
+                log.info('  Manifest List: %s', listinfo['mldigest'])
+                if tags:
```
twaugh: This is testing `tags` but then uses `mltags`. Is that intentional?

dockpulp/cli.py
```diff
@@ -451,6 +451,151 @@ def do_imageids(bopts, bargs, parser):
...
+        if tag in tags:
+            # Does it reference this manifest?
+            if tags[tag] == manifest:
+                active_marker = ' (active)'
+
+        output[layers][manifest]['active'] = active_marker
+
+    # Print out all manifest lists and associated manifests and layers
+    if showlists:
+        # mloutput groups manifest lists by manifest tuples for simple printing
```
twaugh: Why not just iterate over manifest_lists? e.g.
```
for manifest_list in manifest_lists:
    manifest_list_info = repo['manifest_lists'][manifest_list]
    log.info('')
    log.info('  Manifest List: %s', listinfo['mldigest'])
    mltags = ', '.join(manifest_list_info]['tags']
    if mltags:
        log.info('    Tags: %s', mltags)
    for mdigests in manifest_list_info['mdigests']:
        seenmanifests[manifest] = True
        layers = manifest_layers[manifest]
        seenlayers[layers] = True
        _print_manifest_metadata(output[layers], manifest, showschema)
        ... and blobs etc ...
```

dockpulp/__init__.py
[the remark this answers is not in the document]
twaugh: OK, fair enough. Only suggestion I'd have otherwise is to add a comment in the code explaining this.

PR #164, opened 2018-01-31
Added rel-url for docker_rsync_distributor

breillyr:


dockpulp/cli.py
```diff
@@ -378,7 +378,7 @@ def do_create(bopts, bargs, parser):
 
     p.createRepo(repoid, url, desc=opts.description, title=opts.title, protected=opts.protected,
                  productline=productid, library=opts.library, distribution=opts.distribution,
-                 prefix_with=prefix_with)
+                 prefix_with=prefix_with, rel_url=url[1:])
```
twaugh: This looks a little fragile, because it depends on a previous check (that url begins with a string which starts with '/') but the connection is not clear. If that prior check is ever changed or removed, this line will misbehave but that isn't obvious from looking at the code.

Maybe something like this would be better?
```
assert url.startswith('/')
rel_url = url[1:]  # remove leading '/'
p.createRepo(..., rel_url=rel_url)
```

dockpulp/__init__.py
```diff
@@ -1265,6 +1265,11 @@ def createRepo(self, repo_id, url, registry_id=None, desc=None, title=None, prot
                         distributor['distributor_config']['protected'] = protected
                         distributor['distributor_config']['repo-registry-id'] = registry_id
                         distributor['distributor_config']['redirect-url'] = rurl
+                    elif distributor['distributor_type_id'] == 'docker_rsync_distributor':
+                        if rel_url and rurl:
+                            distributor['distributor_config']['repo_relative_path'] = rel_url
+                        elif not rurl:
+                            distributor['distributor_config']['repo_relative_path'] = repo_id
```
twaugh: It is really hard to follow the logic here. The variable names don't give much away, and there is no comment explaining what we're trying to achieve.

I think `rel_url` is the relative URL we were given, and `rurl` is a fully-qualified redirect URL. Why do we set repo_relative_path to repo_id when we are not given either? Isn't that default for that distributor type?

breillyr: Aha, you are correct, I misunderstood what was requested. This was originally to check if the redirect-url was given, and if it was not it would set the repo_relative_path to the repo-id, but since that is what it defaults to there is no need.

dockpulp/cli.py
```diff
@@ -924,8 +925,6 @@ def do_update(bopts, bargs, parser):
             updates['redirect-url'] = opts.redirect
             if opts.redirect.find('/content/') != -1:
                 updates['rel-url'] = opts.redirect.find[opts.redirect.find('content/'):]
-        else:
```
breillyr: Also removed this statement as we do not need to set repoid - it is already the default for the distributor

dockpulp/__init__.py
```diff
@@ -2005,6 +2009,9 @@ def updateRepo(self, rid, update):
                 delta['delta']['notes']['signatures'] = sig
         for distributorkey in distributorkeys:
             delta['distributor_configs'][distributorkey] = {}
+            if distributorkey == rsyncdist and 'rel-url' in update:
+                delta['distributor_configs'][distributorkey]['repo_relative_path'] \
+                    = update['rel-url']
```
twaugh: Minor, but how about this?:
```
config = {}
if distributorkey == rsyncdist and 'rel-url' in update:
    config['repo_relative_path'] = update['rel-url']
delta['distributor_configs'][distributorkey] = config
```
Fits better width-wise so is easier to read, and has fewer lookups.

dockpulp/__init__.py
```diff
@@ -1265,6 +1265,9 @@ def createRepo(self, repo_id, url, registry_id=None, desc=None, title=None, prot
                         distributor['distributor_config']['protected'] = protected
                         distributor['distributor_config']['repo-registry-id'] = registry_id
                         distributor['distributor_config']['redirect-url'] = rurl
+                    elif distributor['distributor_type_id'] == 'docker_rsync_distributor':
+                        if rel_url:
+                            distributor['distributor_config']['repo_relative_path'] = rel_url
```
twaugh: This doesn't get unit test coverage according to:
```
coverage run /usr/bin/py.test tests
coverage html
xdg-open htmlcov/dockpulp___init___py.html
```

dockpulp/__init__.py
```diff
@@ -2005,6 +2009,9 @@ def updateRepo(self, rid, update):
                 delta['delta']['notes']['signatures'] = sig
         for distributorkey in distributorkeys:
             delta['distributor_configs'][distributorkey] = {}
+            if distributorkey == rsyncdist and 'rel-url' in update:
+                delta['distributor_configs'][distributorkey]['repo_relative_path'] \
```
twaugh: This block doesn't get coverage during unit tests.

dockpulp/__init__.py
```diff
@@ -1950,20 +1953,22 @@ def updateRepo(self, rid, update):
...
         webdist = 'docker_distributor_web'
         exportdist = 'docker_distributor_export'
+        rsyncdist = 'docker_rsync_distributor'
 
         disturl = '/pulp/api/v2/repositories/%s/distributors/' % rid
         log.debug("calling %s", disturl)
         blob = self._get(disturl)
         for did in blob:
-            if did['distributor_type_id'] == webdist or did['distributor_type_id'] == exportdist:
+            if did['distributor_type_id'] in (webdist, exportdist):
```
twaugh: Minor: this looks OK, but I don't think we ever test it out properly in unit tests. The 'true' path is never taken. Looking at that `validdistributorkeys` and trying to work out what it does, it seems to me like we may as well set `delta['distributor_configs'][did['id']] = {}` right here inside the loop. Maybe I'm missing something.

twaugh:
There are merge conflicts. Please re-base.

dockpulp/__init__.py
breillyr: Before the addition of the docker_rsync distributor only these two distributors ever needed to be updated - setting this here lets us iterate through those distributors below without checking type_id.

PR #165, opened 2018-02-16
Dlservice5041

breillyr:
Added ability to set and list new reponotes value "include_in_download_service"

### DIFF
--- a/dockpulp/__init__.py
+++ b/dockpulp/__init__.py
@@ -1155,7 +1155,8 @@ class Pulp(object):
 
     def createRepo(self, repo_id, url, registry_id=None, desc=None, title=None, protected=False,
                    distributors=True, prefix_with=PREFIX, productline=None, library=False,
-                   distribution=None, repotype=None, importer_type_id=None, rel_url=None):
+                   distribution=None, repotype=None, importer_type_id=None, rel_url=None,
+                   download=None):
         """Create a docker repository in pulp.
 
         id and description are required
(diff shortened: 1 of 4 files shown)


dockpulp/__init__.py
```diff
@@ -1207,6 +1208,12 @@ def createRepo(self, repo_id, url, registry_id=None, desc=None, title=None, prot
         }
         if rel_url:
             stuff['notes']['relative_url'] = rel_url
+        if download is not None:
+            # json boolean (true/false) or "True"/"False"
```
twaugh: Is there a reason to accept so many things for this parameter, rather than it being `None` or a Boolean?

dockpulp/__init__.py
```diff
@@ -1207,6 +1208,12 @@ def createRepo(self, repo_id, url, registry_id=None, desc=None, title=None, prot
...
+            if download:
+                stuff['notes']['include_in_download_service'] = "True"
+            else:
+                stuff['notes']['include_in_download_service'] = "False"
```
twaugh: For this specific case where `download` is either `None`, in which case we won't want to set include_in_download_service, or a Boolean in wihch case we want either "True" or "False", this could be:
```
if download is not None:
    stuff['notes']['include_in_download_service'] = str(download)
```

dockpulp/__init__.py
breillyr: It can't be a boolean as only JSON booleans or the literal strings "True"/"False" are accepted.

I also didn't want the field to be automatically set if it would never be used for that repo.

PR #166, opened 2018-02-26
WIP - Add autopublish update to updateRepo

bfontecc007:
Signed-off-by: Bret Fontecchio <bfontecc@redhat.com>


dockpulp/__init__.py
```diff
@@ -2025,6 +2031,9 @@ def updateRepo(self, rid, update):
             if 'signature' not in update and sig != "":
                 sig = self.getSignature(sig)
                 delta['delta']['notes']['signatures'] = sig
+        if 'auto-publish' in update:
```
breillyr: This is only including the export and web dists. Put this after line 2041 to include all distributors.

dockpulp/__init__.py
```diff
@@ -2030,6 +2036,9 @@ def updateRepo(self, rid, update):
             if distributortype == rsyncdist and 'rel-url' in update:
                 config['repo_relative_path'] = update['rel-url']
             delta['distributor_configs'][distributorid] = config
+            if 'autopublish' in update:
+                for distributor in delta['distributor_configs']:
+                    distributor['auto-publish'] = update['autopublish']
```
breillyr: This should be 'distributor['auto_publish'], not ['auto-publish']

dockpulp/__init__.py
```diff
@@ -2030,6 +2036,9 @@ def updateRepo(self, rid, update):
             if distributortype == rsyncdist and 'rel-url' in update:
                 config['repo_relative_path'] = update['rel-url']
             delta['distributor_configs'][distributorid] = config
+            if 'autopublish' in update:
```
twaugh: Can we make this consistent with the key name required by Pulp?

bfontecc007: sure

dockpulp/__init__.py
```diff
@@ -1549,6 +1549,12 @@ def listRepos(self, repos=None, content=False, history=False, labels=False, stri
             else:
                 r['distributors'] = None
 
+            r['auto_publish_settings'] = None
+            if blob['distributors']:
+                r['auto_publish_settings'] = [{'distributor_id': distributor['id'],
```
breillyr: Since we will only ever be changing / setting auto_publish by *repo* and not by specific distributors, why don't we make this
 `if blob['distributors']:                                                                     
                r['auto_publish'] = blob['distributors'][0]['auto_publish']`

dockpulp/__init__.py
```diff
@@ -1549,6 +1549,12 @@ def listRepos(self, repos=None, content=False, history=False, labels=False, stri
             else:
                 r['distributors'] = None
 
+            r['auto_publish_settings'] = None
```
breillyr: If you make the change below, make sure to change this to r['auto_publish'] = None

dockpulp/__init__.py
twaugh: Or maybe `... = any(distributor['auto_publish'] for distributor in blob['distributors'])` ? So that it is True if any of the distributors have auto_publish set?

(Do we know for sure the 'auto_publish' key will be set, or should be we using `.get('auto_publish', False)`?)

Also, is `blob['distributors']` always an iterable? If it is, we don't need the `if blob['distributors']` if we iterate over it to get the answer.

breillyr: auto_publish key should always be set as it's hardcoded into /etc/dockpulpdistributors.conf

dockpulp/__init__.py
```diff
@@ -2038,6 +2044,19 @@ def updateRepo(self, rid, update):
...
+                self.updateAutoPublish(rid, did['id'], update['auto_publish'])
+
+    def updateAutoPublish(self, rid, dist_id, auto_publish):
+        d = {
+            "distributor_config": {},
+            "delta": {
+                "auto_publish": auto_publish}}
+        tid = self._put('/pulp/api/v2/repositories/%s/distributors/%s/' % (rid, dist_id),
+                        data=json.dumps(d))
+        self.watch(tid)
```
twaugh: Let's not wait for each task before starting the next one. Instead, this should return the tid and the caller can wait for all of them together.

lcarva: Good idea! Maybe this method can take a `watch=True` parameter so if it's used on its own it does wait for task. `updateRepo` should invoke it with `watch=False`. In either case, method should return task ID. Thoughts?

twaugh: I like it.

breillyr: Sounds good to me.

dockpulp/__init__.py
```diff
@@ -2038,6 +2038,25 @@ def updateRepo(self, rid, update):
                         data=json.dumps(delta))
         self.watch(tid)
 
+        if 'auto_publish' in update:
+            dist_tids = []
+            for did in blob:
+                dist_tids.append(self.updateAutoPublish(rid, did['id'],
+                                 update['auto_publish'], False))
```
lcarva: Let's call out `watch=False` so it's clearer what the bool is for.

bfontecc007: sure

dockpulp/cli.py
```diff
@@ -935,6 +935,8 @@ def do_update(bopts, bargs, parser):
     parser.add_option('-t', '--title', help='set the title (short desc)')
     parser.add_option('-s', '--signature', help='set the signatures field for the repo')
     parser.add_option('--distribution', help='set the distribution field for the repo')
+    parser.add_option('-a', '--autopublish', help='set the autopublish bit for the repo. '
```
lcarva: This should probably be `--auto-publish` (and "auto publish" in `help`) since we treat it as two different words in other places.

dockpulp/__init__.py
```diff
@@ -2038,6 +2038,25 @@ def updateRepo(self, rid, update):
                         data=json.dumps(delta))
         self.watch(tid)
 
+        if 'auto_publish' in update:
+            dist_tids = []
+            for did in blob:
+                dist_tids.append(self.updateAutoPublish(rid, did['id'],
+                                 update['auto_publish'], watch=False))
```
twaugh: Indentation doesn't align here. `update[...]` is an argument to the `self.updateAutoPublish()` method call so I'd expect it to align after the `(` of that method call, not the `append()` method call.

dockpulp/cli.py
```diff
@@ -711,6 +711,27 @@ def do_json(bopts, bargs, parser):
...
+    opts, args = parser.parse_args(bargs)
+    p = pulp_login(bopts)
+
+    log.removeHandler(sh)
+    log.addHandler(dockpulp.NullHandler())
+
+    if len(args) == 0:
+        raise RuntimeError('Specify a repo')
+
+    repos = p.getRepos(args, fields="id", distributors=True)
```
lcarva: Let's either remove this function, or make sure it takes enough parameters so we don't have to hard code them.

dockpulp/__init__.py
```diff
@@ -2038,6 +2041,25 @@ def updateRepo(self, rid, update):
                         data=json.dumps(delta))
         self.watch(tid)
```
twaugh: Any reason to wait for this task to finish before issuing the distributor changes? i.e. can we wait for this `tid` along with all the other `dist_tids`?

lcarva: I wasn't sure if pulp would do the right thing.

dockpulp/cli.py
lcarva: Any updates on this?

dockpulp/__init__.py
breillyr: Agree with @lcarva, unsure if there's the possibility of something going wrong if distributors were updated in a strange order

lcarva:
Missing unit tests.

dockpulp/cli.py
bfontecc007: removing it

twaugh:
Looks great, but has merge conflicts. Please re-base.

tests/test_pulp_instance.py
```diff
@@ -323,11 +323,13 @@ def test_getPrefix(self, pulp):
     ])
     @pytest.mark.parametrize('dist', [None, 'beta'])
     @pytest.mark.parametrize('download', [True, False, None])
-    def test_updateRedirect(self, pulp, rid, redirect, dist, download):
-        update = {'redirect-url': redirect, 'rel-url': 'content/'}
+    @pytest.mark.parametrize('auto_publish', [True, False])
+    def test_updateRedirect(self, pulp, rid, redirect, dist, download, auto_publish):
```
lcarva: It seems odd to add auto_publish to `test_updateRedirect`. Should this be its own test?

bfontecc007: It can be. I just figured it would save some space in the code

PR #170, opened 2018-03-02
Output tasks json in silent mode

lcarva:


lcarva:
@twaugh, didn't realize we actually had unit tests for CLI; added.

PR #171, opened 2018-03-02
Fixed bug with schema 2 history output

breillyr:


lcarva:
Can this be unit tested?

dockpulp/__init__.py
```diff
@@ -1699,11 +1699,18 @@ def listRepos(self, repos=None, content=False, history=False, labels=False, stri
                         clean.sort()
                         continue
                     for manifest in r['manifests'].keys():
+                        sver = r['manifests'][manifest]['schema_version']
+                        if sver:
+                            manifesturl = '/pulp/docker/v2/%s/manifests/%s/%s' % (
+                                blob['id'], sver, manifest)
```
twaugh: Do we really want to hit the data store directly, rather than using the Docker Registry HTTP API offered by Crane?

Doing it this way is fragile for exactly the reason we're seeing here: Pulp can change the layout of this at any time.

dockpulp/__init__.py
```diff
@@ -1699,11 +1699,18 @@ def listRepos(self, repos=None, content=False, history=False, labels=False, stri
                         clean.sort()
                         continue
                     for manifest in r['manifests'].keys():
+                        sver = r['manifests'][manifest]['schema_version']
```
twaugh: Is the presence of this what determines which path to try? I thought it might be whether or not that repository has been re-published since upgrade, in which case we'd need to try the 'new' form, then fall back to the 'old' form on 404(?).

breillyr: schema version just determines if it's a '/1/manifest' or '/2/manifest'. I'll have both of them fall back to the 'old' form though.

twaugh: Couldn't there be a case where 'schema_version' is present (indicating it is Pulp >= 2.15), but the repository has not been published since an upgrade from Pulp < 2.15? In that case, the presence of 'schema_version' doesn't tell us anything about whether to try to the old or new path.

Also, if this runs against Pulp < 2.15 and 'schema_version' is not present, won't this look-up cause a traceback? (i.e. do we want to use `.get()`?)

dockpulp/__init__.py
```diff
@@ -1699,15 +1699,29 @@ def listRepos(self, repos=None, content=False, history=False, labels=False, stri
...
+                        if sver:
+                            manifesturl = '/pulp/docker/v2/%s/manifests/%s/%s' % (
+                                blob['id'], sver, manifest)
+                        else:
+                            manifesturl = '/pulp/docker/v2/%s/manifests/%s' % (
```
twaugh: Can we define these paths once?

dockpulp/__init__.py
```diff
@@ -1699,15 +1699,29 @@ def listRepos(self, repos=None, content=False, history=False, labels=False, stri
...
+                                blob['id'], manifest)
+
                         try:
-                            data = self._get('/pulp/docker/v2/%s/manifests/%s' % (
-                                blob['id'], manifest))
+                            data = self._get(manifesturl)
                         except errors.DockPulpError:
```
twaugh: Do we know what status code we expect in the fall-back case? Should it be not_found?

I doubt we want to just fall back on *any* error.

dockpulp/__init__.py
breillyr: If the manifest has a schema_version but uses the old location, the 'fall back' call should cover that case.
And it seems as if a manifest has the new location it must have a schema_version as well, as the location corresponds to /1/ or /2/. Is that sufficient coverage?

dockpulp/__init__.py
breillyr: We expect 404, unfortunately there is no specific dockpulp error for status codes - it just prints out the status code as part of the error message. Seems like a good RFE to add more specific error handling here.

dockpulp/__init__.py
```diff
@@ -1699,15 +1699,27 @@ def listRepos(self, repos=None, content=False, history=False, labels=False, stri
                         clean.sort()
                         continue
                     for manifest in r['manifests'].keys():
+                        sver = r['manifests'][manifest].get('schema_version', None)
+                        if sver:
+                            manifestpath = '%s/%s' % (sver, manifest)
```
twaugh: There are lots of lines not covered by unit tests. We can reduce that, and reduce repetition, by re-writing this slightly:
```
manifestpaths = [manifest]
if sver:
    # Try new path first, then fall back to old path
    manifestpaths.insert(0, '%s/%s' % (sver, manifest))
last_attempt = len(manifestpaths) - 1
for attempt, manifestpath in enumerate(manifestpaths):
    try:
        data = self._get(... % manifestpath)
    except errors.DockPulpError:
        # Any more to try?
        if attempt < last_attempt:
            continue
        log.warning(...)
        ...
```
What do you think?

dockpulp/__init__.py
```diff
@@ -1699,15 +1699,27 @@ def listRepos(self, repos=None, content=False, history=False, labels=False, stri
...
+                            except errors.DockPulpError:
+                                # Any more to try?
+                                if attempt < last_attempt:
+                                    continue
+                                log.warning("Manifest history info unreachable, skipping %s",
+                                            manifest)
+                                r['manifests'][manifest]['v1parent'] = None
+                                r['manifests'][manifest]['v1id'] = None
+                                r['manifests'][manifest]['v1labels'] = None
+                                continue
```
lcarva: This will `continue` the `for attempt, manifestpath in enumerate(manifestpaths):` loop. Is that what you want? Previous code had it on `for manifest in r['manifests'].keys():` loop.

dockpulp/__init__.py
```diff
@@ -1699,15 +1699,27 @@ def listRepos(self, repos=None, content=False, history=False, labels=False, stri
...
+                            manifestpaths.insert(0, '%s/%s' % (sver, manifest))
+                        last_attempt = len(manifestpaths) - 1
+                        for attempt, manifestpath in enumerate(manifestpaths):
+                            try:
+                                data = self._get('/pulp/docker/v2/%s/manifests/%s' % (
+                                    blob['id'], manifestpath))
+                                break
+                            except errors.DockPulpError:
+                                # Any more to try?
+                                if attempt < last_attempt:
```
lcarva: I never thought I'd say this, but for-else might actually be cleaner here:

```python
for manifestpath in manifestpaths:
  try:
    data = self._get('/pulp/docker/v2/%s/manifests/%s' % (blob['id'], manifestpath))
    break
  except errors.DockPulpError:
    pass
else:
  log.warning("Manifest history info unreachable, skipping %s", manifest)
  r['manifests'][manifest]['v1parent'] = None
  r['manifests'][manifest]['v1id'] = None
  r['manifests'][manifest]['v1labels'] = None
  continue
```

You can also use a `found` boolean if for-else is not your cup of tea.

This also makes it clearer that `continue` is meant to apply to the correct for-loop.

dockpulp/__init__.py
rcerven: if there is schema version we have 2 options, if we have still attempts to left we can continue

lcarva: right, but that's not what the original code was doing. If it couldn't find manifest info, it carried on to the next manifest.

PR #173, opened 2018-03-20
Deprecating HIDDEN repo and replacing with 'origin' repos

breillyr:
Replaced all instances of HIDDEN repo dependence with origin repos based on each existing repo.
If the origin repo does not exist it will be automatically created during those operations.

### DIFF
--- a/dockpulp/__init__.py
+++ b/dockpulp/__init__.py
@@ -67,6 +67,7 @@ DEFAULT_CONFIG_FILE = '/etc/dockpulp.conf'
 DEFAULT_DISTRIBUTORS_FILE = '/etc/dockpulpdistributors.json'
 DEFAULT_DISTRIBUTIONS_FILE = '/etc/dockpulpdistributions.json'
 PREFIX = 'redhat-'
+ORIGIN_PREFIX = 'origin-'
 
 
 # Setup our logger
@@ -965,6 +966,27 @@ class Pulp(object):
             new_repos.append(new_repo_key)
         return new_repos
 
+    def _find_content_source(self, content_id, content_filter, content_types):
+        # content_filter: 'digest' or 'image_id'
+        # content_types: [V2_C_TYPE, V2_LIST, V2_BLOB] or [V1_C_TYPE]
+
+        search_data = json.dumps({
+            'criteria': {
+                'filters': {
+                    content_filter: content_id
+                },
+            },
+            'include_repos': True
+        })
+        for content_type in content_types:
+            result = self._post('/pulp/api/v2/content/units/%s/search/' % content_type,
+                                data=search_data)
+            if result:
+                source = result[0]['repository_memberships'][0]
+                return source
+
+        raise errors.DockPulpError('Image cannot be found in Pulp: %s' % content_id)
+
     # public methods start here, alphabetically
 
     def associate(self, dist_id, repo, type_id=None):
@@ -1040,14 +1062,17 @@ class Pulp(object):
         for upload in uploads:
             self._deleteUploadRequest(upload)
 
-    def copy(self, drepo, img, source=HIDDEN):
+    def copy(self, drepo, img, source=None):
         """Copy an image from one repo to another."""
         if img.startswith("sha256:"):
+            content_types = [V2_C_TYPE, V2_LIST, V2_BLOB]
+            if source is None:
+                source = self._find_content_source(img, 'digest', content_types)
 
             data = {
                 'source_repo_id': source,
                 'criteria': {
-                    'type_ids': [V2_C_TYPE, V2_BLOB, V2_TAG, V2_LIST],
+                    'type_ids': content_types,
                     'filters': {
                         'unit': {
                             "$or": [{'digest': img}, {'manifest_digest': img}]
@@ -1058,11 +1083,14 @@ class Pulp(object):
             }
 
         else:
+            content_types = [V1_C_TYPE]
+            if source is None:
+                source = self._find_content_source(img, 'image_id', content_types)
 
             data = {
                 'source_repo_id': source,
                 'criteria': {
-                    'type_ids': [V1_C_TYPE],
+                    'type_ids': content_types,
                     'filters': {
                         'unit': {
                             'image_id': img
@@ -1153,10 +1181,20 @@ class Pulp(object):
         else:
             return tasks
 
+    def createOriginRepo(self, repo_id):
+        if not repo_id.startswith(ORIGIN_PREFIX):
+            repo_id = ORIGIN_PREFIX + repo_id
+        try:
+            self.listRepos(repos=[repo_id])
+        except errors.DockPulpError:
+            # only create origin repo if it does not exist
+            self.createRepo(repo_id, None, distributors=False, prefix_with=ORIGIN_PREFIX,
+                            is_origin=True)
+
     def createRepo(self, repo_id, url, registry_id=None, desc=None, title=None, protected=False,
                    distributors=True, prefix_with=PREFIX, productline=None, library=False,
                    distribution=None, repotype=None, importer_type_id=None, rel_url=None,
-                   download=None):
+                   download=None, is_origin=False):
         """Create a docker repository in pulp.
 
         id and description are required
@@ -1246,7 +1284,7 @@ class Pulp(object):
                 sig = self.getSignature(distconf['signature'])
                 stuff['notes']['signatures'] = sig
             stuff['notes']['distribution'] = distribution
-        elif self.dists and repo_id != HIDDEN and repo_id != SIGSTORE:
+        elif self.dists and repo_id != HIDDEN and repo_id != SIGSTORE and not is_origin:
             raise errors.DockPulpError("Env %s requires distribution defined at repo creation" %
                                        self.env)
         if repotype:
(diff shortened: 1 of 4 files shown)


dockpulp/__init__.py
```diff
@@ -1961,16 +1981,18 @@ def syncRepo(self, env=None, repo=None, config_file=DEFAULT_CONFIG_FILE,
         manifests = list(set(newmanifests) - set(oldmanifests))
         manifests.sort()
 
-        # Need to maintain HIDDEN
-        if repo != HIDDEN and (imgs or manifests):
+        # Need to maintain origin repo
+        self.createOriginRepo(origin_repo)
+
+        if imgs or manifests:
```
midnightercz: this is not related to this PR. But for schema2 you will also have to check for manifest lists and tags. Then it makes sense to copy everything everytime and leave filtering of already copied units to pulp.

dockpulp/__init__.py
```diff
@@ -67,6 +67,7 @@
 DEFAULT_DISTRIBUTORS_FILE = '/etc/dockpulpdistributors.json'
 DEFAULT_DISTRIBUTIONS_FILE = '/etc/dockpulpdistributions.json'
 PREFIX = 'redhat-'
+ORIGIN = 'origin-'
```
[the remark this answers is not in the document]
midnightercz: also, we internally use temp- instead of origin-

dockpulp/__init__.py
```diff
@@ -1153,10 +1154,17 @@ def crane(self, repos=[], wait=True, skip=False, force_refresh=False):
         else:
             return tasks
 
+    def createOriginRepo(self, repo_id):
+        try:
+            self.listRepos(repos=repo_id)
+        except errors.DockPulpError:
+            # only create origin repo if it does not exist
+            self.createRepo(repo_id, None, distributors=False, prefix_with=ORIGIN, origin=True)
```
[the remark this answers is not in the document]
midnightercz: If we consider purpose of origin repos e.i. avoid have orphans in pulp db, old content is available in HIDDEN. Only new content will be in origin- repos. So it's probably not necessary to copy everything over.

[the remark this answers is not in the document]
lcarva: Good point. Getting rid of the existing HIDDEN repo can be done at a later time if that's desired.

dockpulp/__init__.py
```diff
@@ -1040,10 +1041,28 @@ def cleanUploadRequests(self):
         for upload in uploads:
             self._deleteUploadRequest(upload)
 
-    def copy(self, drepo, img, source=HIDDEN):
+    def copy(self, drepo, img, source=None):
         """Copy an image from one repo to another."""
         if img.startswith("sha256:"):
 
+            if source is None:
+                search_data = json.dumps({
```
lcarva: It would be really neat if this was a helper method:
```python
def _find_content_source(content_id, content_filter, content_types):
  # content_filter: 'digest' or 'image_id'
  # content_types: (V2_C_TYPE, V2_LIST, V2_BLOB, V2_TAG) or (V1_C_TYPE,)
``` 

It would make it much easier to read this method.

dockpulp/__init__.py
```diff
@@ -1153,10 +1188,18 @@ def crane(self, repos=[], wait=True, skip=False, force_refresh=False):
         else:
             return tasks
 
+    def createOriginRepo(self, repo_id):
+        try:
+            self.listRepos(repos=repo_id)
```
lcarva: Should prefix be added here before querying it? `self.listRepos(repos=ORIGIN_PREFIX + repo_id)`
Or better, do it conditionally if it's not already done. I say this because this method is public and having the caller add prefix seems to be an odd requirement.

dockpulp/__init__.py
```diff
@@ -1354,31 +1403,30 @@ def getAllRepoIDs(self):
...
+            img = self._post('/pulp/api/v2/content/units/%s/search/' % V1_C_TYPE,
                              data=json.dumps(data))[0]
         except IndexError:
             log.info('missing parent layer %s', iid)
             log.info('skipping layer')
             return parents
-        img['metadata'].setdefault('parent_id', None)
-        par = img['metadata']['parent_id']
+        img.setdefault('parent_id', None)
+        par = img['parent_id']
```
lcarva: `setdefault` actually returns the value, this could be `par = img.setdefault('parent_id', None)`

dockpulp/__init__.py
```diff
@@ -1953,6 +2004,11 @@ def syncRepo(self, env=None, repo=None, config_file=DEFAULT_CONFIG_FILE,
         else:
             oldmanifests = repoinfo['manifests'].keys()
 
+        if len(repoinfo['manifest_lists'].keys()) == 0:
```
lcarva: This is not needed. `dict.keys()` will return an empty list. `oldmanifestlists = repoinfo['manifests'].keys()` should handle both cases.

dockpulp/__init__.py
```diff
@@ -2127,15 +2184,24 @@ def upload(self, image):
                 'filename': os.path.basename(image)
             }
         }
-        log.info('adding %s to %s' % (iid, HIDDEN))
+        if drepo != HIDDEN and not drepo.startswith(ORIGIN_PREFIX):
+            origin_drepo = self.getOriginPrefix() + drepo
+        else:
+            origin_drepo = drepo
+        self.createOriginRepo(drepo)
```
lcarva: Wouldn't this create an origin repo for hidden repo in some cases?

dockpulp/__init__.py
```diff
@@ -2127,15 +2184,24 @@ def upload(self, image):
...
-            '/pulp/api/v2/repositories/%s/actions/import_upload/' % HIDDEN,
+            '/pulp/api/v2/repositories/%s/actions/import_upload/' % origin_drepo,
             data=json.dumps(data))
         timer = max(self.timeout, (size / mb) * 2)  # wait 2 seconds per megabyte, or timeout,
         self.watch(tid, timeout=timer)  # whichever is greater
         self._deleteUploadRequest(rid)
+        # use filter to copy all new images
+        pulp_filter = {'unit': {
+            '$or': [{'image_id': img} for img in newimgs]}}
+        self.copy(drepo, source=origin_drepo, filters=pulp_filter, v1=True, v2=False)
```
lcarva: `copy_filters` ?

dockpulp/__init__.py
```diff
@@ -2085,13 +2140,15 @@ def updateAutoPublish(self, rid, dist_id, auto_publish, watch=True):
             self.watch(tid)
         return tid
 
-    def upload(self, image):
+    def upload(self, image, drepo=HIDDEN):
```
lcarva: I think this is the only method that we can't automatically transition to origin repo... 
I'm wondering if this should be a breaking change, i.e. always require `drepo` parameter.

Although, I guess it's not too bad that it's still using `HIDDEN`.

dockpulp/__init__.py
rcerven: it seems to be query for any repo, even without origin, and crates origin only if repo doesn't exists

dockpulp/__init__.py
```diff
@@ -1294,6 +1337,12 @@ def createRepo(self, repo_id, url, registry_id=None, desc=None, title=None, prot
                     continue
         else:
             stuff['distributors'] = []
+
+        if not is_origin and repo_id != HIDDEN and repo_id != SIGSTORE:
+            # want to create origin- repo for every new repo
+            # do this at the end in case of errors
+            self.createOriginRepo(ORIGIN_PREFIX + repo_id)
```
rcerven: can't we just raise error in createOriginRepo, and not call createRepo again,
so it if doesn't raise it can just continue with creation?

dockpulp/cli.py
```diff
@@ -874,6 +873,15 @@ def do_sync(bopts, bargs, parser):
             log.info(manifest)
     log.info('')
 
+    log.info('synced manifest lists:')
+
+    if len(manifest_lists) == 0:
```
rcerven: can be just "if manifest_lists" since it will be empty list if it is empty

dockpulp/__init__.py
breillyr: It wouldn't have if createOriginRepo did not automatically prepend w/ the origin prefix (as it would have checked if 'redhat-everything' exists and then stopped there), but now that it does this must be changed, yes.

dockpulp/__init__.py
breillyr: I'd rather not make it a breaking change as I don't think there is much of a need. Other tools will have to be updated to match this and I don't want the functionality to break in the meantime. Once they're updated I can remove it at that point and require repo input.

dockpulp/__init__.py
breillyr: What we want to avoid here is the origin repo being successfully created while the actual repo fails - the error the comment is talking about there is a parsing error for the original repo. Essentially it's the sanity check before we commit to creating the origin repo - if we fail before we even make the call to Pulp, the origin repo should not be created with that faulty information.

dockpulp/__init__.py
```diff
@@ -1058,11 +1083,14 @@ def copy(self, drepo, img, source=HIDDEN):
             }
 
         else:
+            content_types = [V1_C_TYPE]
```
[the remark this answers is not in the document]
lcarva: Isn't `docker_tag` v2 content that is automatically generated for manifests?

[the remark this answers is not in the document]
twaugh: Well sure, but you don't reference a docker_tag unit by the digest it refers to.

PR #174, opened 2018-04-06
S2tags 5368

breillyr:


dockpulp/cli.py
```diff
@@ -511,14 +511,15 @@ def _print_v2_images(repo, showlists, justmanifests, showhistory, showlabels, sh
...
         # Is there a docker_tag unit for this name?
-        if tag in tags:
-            # Does it reference this manifest?
-            if tags[tag] == manifest:
-                active_marker = ' (active)'
+        for tag in taglist:
+            if tag in tags:
+                # Does it reference this manifest?
+                if tags[tag] == manifest:
+                    active_marker = ' (active)'
```
lcarva: Do we need to `break` after we find it?

lcarva:
It would be great to see an example output.

PR #181, opened 2018-07-06
syncRepo: copy new files with an explicit filter

mlangsdorf:
syncRepo: copy new files with an explicit filter
    
apparently pulp does not do a good job of filtering out old files
when copying them to a new repo. In syncRepo, add a filter containing
the new images, new manifests, and new manifest lists.

Signed-off-by: Mark Langsdorf <mlangsdo@redhat.com>

### DIFF
--- a/dockpulp/__init__.py
+++ b/dockpulp/__init__.py
@@ -2003,8 +2003,19 @@ class Pulp(object):
         # Need to maintain origin repo
         self.createOriginRepo(origin_repo)
 
-        # copy with no filter. Pulp will handle already copied units automatically
-        self.copy_filters(origin_repo, repo)
+        # copy with filter to only select the new units
+        new_units = []
+        filters = {}
+        if imgs:
+            img_ids = {"image_id": {"$in": imgs}}
+            new_units.append(img_ids)
+        if manifests or manifest_lists:
+            digests = [manifest for manifest in manifests]
+            digests.extend([manifest_list for manifest_list in manifest_lists])
+            new_units.append({"manifest_digest": {"$in": digests}})
+        if new_units:
+            filters["unit"] = {"$or": new_units}
+        self.copy_filters(origin_repo, repo, filters)
 
         return (imgs, manifests, manifest_lists)
 
--- a/tests/test_pulp_instance.py
+++ b/tests/test_pulp_instance.py
@@ -785,12 +785,34 @@ class TestPulp(object):
             .and_return(None))
         pulp.remove_filters(repo)
 
-    def test_sync(self, pulp):
+    @pytest.mark.parametrize(('new_images', 'image_results'), [
+        ([], []),
+        (set(['foobar']), ['foobar']),
+        (set(['foobar', 'barfoo']), ['barfoo', 'foobar']),
+    ])
+    @pytest.mark.parametrize(('new_manifests', 'new_manifest_lists', 'manifest_filter',
+                              'manifest_results', 'manifest_list_results'), [
+        ({'123456': 'foobar'}, {}, None, [], []),
+        ({'123456': 'foobar', '567890': 'latest'}, {}, ['567890'], ['567890'], []),
+        ({'923412': 'barfoo', '123456': 'foobar', '567890': 'latest'}, {},
+         ['567890', '923412'], ['567890', '923412'], []),
+        ({}, {'123456': 'foobar'}, ['123456'], [], ['123456']),
+        ({}, {'923412': 'barfoo', '123456': 'foobar'}, ['123456', '923412'],
+         [], ['123456', '923412']),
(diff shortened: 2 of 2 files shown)


dockpulp/__init__.py
```diff
@@ -2003,8 +2003,12 @@ def syncRepo(self, env=None, repo=None, config_file=DEFAULT_CONFIG_FILE,
...
         self.createOriginRepo(origin_repo)
 
-        # copy with no filter. Pulp will handle already copied units automatically
-        self.copy_filters(origin_repo, repo)
+        # copy by hand. Pulp should handle already copied units automatically, but doesn't.
+        for collection in (imgs, manifests, manifest_lists):
+            for item in collection:
+                self.copy(origin_repo, item, repo)
+
+        # self.copy_filters(origin_repo, repo)
```
twaugh: I think we want to submit the copy as a single request rather than multiple requests each for a single unit.
Currently (https://github.com/mlangsdorf/dockpulp/blob/376f4dfa5b2a583f9121e33de4abb629dc77c0fd/dockpulp/__init__.py#L1112-L1119) we're submitting this request:
```
{
  "source_repo_id": "...",
  "criteria": {
    "type_ids": ["docker_image", "docker_manifest", "docker_blob", "docker_tag", "docker_manifest_list"],
    "filters": {},
  },
  "override_config": {}
}
```

Instead I think we want to use an actual filter, like this:
```
{
  "source_repo_id": "...",
  "criteria": {
    "type_ids": ["docker_image", "docker_manifest", "docker_blob", "docker_tag", "docker_manifest_list"],
    "filters": {
      "unit": {
        "$or": [
          {"image_id": {"$in": ["123456", "234567", "... image IDs ..."]}},
          {"manifest_digest": {"$in": ["sha256:012345", "sha256:45462", "... image manifests and manifest lists ..."]}}
        ]
      }
    },
  },
  "override_config": {}
}
```

mlangsdorf: what's the filter syntax for manifest_lists?

twaugh: The "manifest_digest" property is common to both image manifests and manifest lists, so you can include both sets of digests in the "$in" match list.

mlangsdorf: Fixed, thanks.

tests/test_pulp_instance.py
```diff
@@ -785,12 +785,34 @@ def test_remove_filters(self, pulp):
             .and_return(None))
         pulp.remove_filters(repo)
 
-    def test_sync(self, pulp):
+    @pytest.mark.parametrize(('new_images', 'image_results'), [
+        ({}, []),
+        ({'foobar'}, ['foobar']),
```
twaugh: For Python 2.6 a set literal needs to be written as `set(['foobar'])`. This is causing tests to fail.

dockpulp/__init__.py
```diff
@@ -2003,8 +2003,21 @@ def syncRepo(self, env=None, repo=None, config_file=DEFAULT_CONFIG_FILE,
...
-        # copy with no filter. Pulp will handle already copied units automatically
-        self.copy_filters(origin_repo, repo)
+        # copy with filter to only select the new units
+        new_units = []
+        if imgs:
+            img_ids = {"image_id": {"$in": imgs}}
+            new_units.append(img_ids)
+        if manifests or manifest_lists:
+            digests = ["sha256:" + manifest for manifest in manifests]
+            digests.extend(["sha256:" + manifest_list for manifest_list in manifest_lists])
```
twaugh: I feel sure "sha256:" is already part of the digest here.

mlangsdorf: Fixed.

tests/test_pulp_instance.py
mlangsdorf: Fixed, eventually.

dockpulp/__init__.py
```diff
@@ -2003,8 +2003,21 @@ def syncRepo(self, env=None, repo=None, config_file=DEFAULT_CONFIG_FILE,
...
+        if manifests or manifest_lists:
+            digests = [manifest for manifest in manifests]
+            digests.extend([manifest_list for manifest_list in manifest_lists])
+            new_units.append({"manifest_digest": {"$in": digests}})
+        filters = {
+            "unit": {
+                "$or": new_units
+            }
+        }
+        self.copy_filters(origin_repo, repo, filters)
```
twaugh: Unless `new_units` is truthy we should avoid construct `filters` and calling `copy_filters()`, to avoid this (seen when sync adds no new units):
```
ERROR     Subtask failed: {u'code': u'PLP0000', u'data': {}, u'description': u'$and/$or/$nor must be a nonempty array', u'sub_errors': []}
```

mlangsdorf: Thanks for testing. Fixed.

PR #185, opened 2018-08-08
Improve sync efficiency

twaugh:
Instead of fetching the entire repository before and after the sync and comparing them, find new units by fetching the repository content after the sync, limiting the search to those units updated since the start of the sync task.

To help with this, add a new 'since' parameter to listRepos(), and make watch() return the task report.

### DIFF
--- a/dockpulp/__init__.py
+++ b/dockpulp/__init__.py
@@ -15,6 +15,7 @@
 
 import atexit
 import ConfigParser
+from datetime import datetime
 import hashlib
 import logging
 import os
@@ -1969,9 +1970,9 @@ class Pulp(object):
             repo = prefix_with + repo
         origin_repo = origin_prefix + repo
 
-        repoinfo = self.listRepos(repo, True)
         if not upstream_name:
-            upstream_name = repoinfo[0]['docker-id']
+            repoinfo = self.listRepos(repos=[repo])[0]
+            upstream_name = repoinfo['docker-id']
 
         if not feed:
             self._getRepo(env, config_file)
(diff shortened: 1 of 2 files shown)


dockpulp/__init__.py
```diff
@@ -1977,32 +1984,30 @@ def syncRepo(self, env=None, repo=None, config_file=DEFAULT_CONFIG_FILE,
...
+        try:
+            dt = datetime.fromisoformat(start_time)
+            start = int(dt.timestamp())
+        except AttributeError:
+            # For Python < 3.7. This is a mess.
+            dt = datetime.strptime(start_time, '%Y-%m-%dT%H:%M:%SZ')
+            epoch = datetime(1970, 1, 1)
+            start = (dt - epoch).total_seconds()
+
+        repoinfo = self.listRepos(repos=[repo], content=True,
```
lcarva: This may pick up content units that have been synced by another task in the case there are multiple OSBS builds running for the same repo (think scratch OSBS builds).

I think that's ok.

dockpulp/__init__.py
```diff
@@ -1977,32 +1985,30 @@ def syncRepo(self, env=None, repo=None, config_file=DEFAULT_CONFIG_FILE,
...
-        repoinfo = repoinfo[0]
+        # Need to maintain origin repo
+        self.createOriginRepo(origin_repo)
 
-        newimgs = set(repoinfo['images'])
-        imgs = list(newimgs - oldimgs)
+        task = self.watch(tid)
+        start_time = task['start_time']
+        try:
+            dt = datetime.fromisoformat(start_time)
```
lcarva: Does it makes sense to move this processing to `listRepo` method? I think it's more intuitive if `listRepo` took a datetime object as `since` and did whatever was needed to turn that into a pulp query.

dockpulp/__init__.py
twaugh: Yes, seems like that isn't a big issue.

dockpulp/__init__.py
twaugh: Good idea, yes.

twaugh:
@lcarva, updated, see what you think.

PR #189, opened 2018-08-17
Paginate repo contents to avoid pulp memory issues

lcarva:

### DIFF
--- a/dockpulp/__init__.py
+++ b/dockpulp/__init__.py
@@ -191,7 +191,7 @@ class Crane(object):
         manifest = manifest.replace('=', ':')
         return (repo, manifest)
 
-    def confirm(self, repos, v1=True, v2=True, silent=True, check_layers=False):
+    def confirm(self, repos, v1=True, v2=True, silent=True, check_layers=False, paginate=True):
 
         auto = 'auto'
 
@@ -199,7 +199,7 @@ class Crane(object):
             v1 = True
             v2 = auto  # auto, based on /v2/ response from crane
 
-        repos = self.p.listRepos(repos=repos, content=True)
+        repos = self.p.listRepos(repos=repos, content=True, paginate=paginate)
         self.errors = 0
         self.errorids = {}
         repoids = {}
@@ -715,7 +715,7 @@ class Crane(object):
 
         return result
 
-    def _test_sigstore(self, signatures, prefix_with=PREFIX, exception=None):
+    def _test_sigstore(self, signatures, prefix_with=PREFIX, exception=None, paginate=True):
         """Confirm we can reach CDN and get data back from it."""
         result = {'error': False, 'sigs_in_pulp_not_crane': [],
                   'manifests_in_sigstore_not_repo': [],
@@ -734,7 +734,8 @@ class Crane(object):
                 log.warning("signature %s is missing '@', skipping", signature)
                 continue
             manifests.setdefault(repo, []).append(manifest)
-        signed_repos = self.p.listRepos(manifests.keys(), content=True, strict=False)
+        signed_repos = self.p.listRepos(manifests.keys(), content=True, strict=False,
+                                        paginate=paginate)
         repo_sigs = {}
         for repo in signed_repos:
             repo_sigs[repo['id']] = repo['signatures']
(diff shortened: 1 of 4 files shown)


lcarva:
@twaugh, @breillyr PTAL

twaugh:
Do we care about the possibility of units we already saw being deleted before our request loop finishes?

lcarva:
> Do we care about the possibility of units we already saw being deleted before our request loop finishes?

I don't know how to get around it.

breillyr:
LGTM and tested locally, works as expected.

twaugh:
> I don't know how to get around it.

Could adjust skip by less than the amount we received and check for expected overlap.

breillyr:
>  Could adjust skip by less than the amount we received and check for expected overlap.

Even then there's no guarantee, a single image could have been removed at the start of the block for example. I'm not entirely sure how Pulp handles these requests without pagination, considering the time required to get responses back on large repos couldn't we have run into that issue already in the current implementation anyways?

lcarva:
The same problem will occur with added content, which happens more frequently. Maybe for that case, we can limit the query by created timestamp?

What if result doesn't match up? Raise an exception?

lcarva:
Also, pulp is quite weird with pagination. It seems to not always respect the limit option.

For example, if there are 40,000 content units in repo, I've seen the following:
1. Request 10,000 records with skip set to 0, pulp returns 10,000 records.
2. Request 10,000 records with skip set to 10,000, pulp returns 7,201 records.
3. Request 10,000 records with skip set to 17,201, pulp returns 10,000 records.
4. And so on until no records are returned.

I don't know why it returns less than what's requested in this case. 

Is paginating results potentially a dangerous change?

twaugh:
I was thinking that if the overlap doesn't match up we'd decrease skip until we do see an overlap.

Maybe it should be possible to disable pagination for the dockpulp API call?

lcarva:
> Maybe it should be possible to disable pagination for the dockpulp API call?

That makes sense. Default to enabled.

To summarize:

1. Pagination is optional - on by default
2. Query should use a timestamp to avoid picking up units added between page requests
3. Query should check for overlap to handle units deleted between page requests

For third item, I'm wondering if we should restart the whole query if overlap check fails. My concern with increasing overlap by one and retrying is that we may get stuck in a loop where we're hitting the server hundreds of times (think garbage collection). Restarting the whole query should only be attempted 3(?) times before dock-pulp fails out.

Thoughts?

twaugh:
Yes, let's restart the query a limited number of times. That all looks good.

breillyr:
Looks good to me.

lcarva:
I can't get a successful run with the overlap check because assertion seems to randomly fail.
Tried explicit sorting but that didn't change the outcome. Pushed commit so you can see where I'm at.

lcarva:
It appears that the overlap check fails in the case where pulp returns less items than were requested as I mentioned in https://github.com/release-engineering/dockpulp/pull/189#issuecomment-414686067.
Digging further.

lcarva:
Alright, I figured out the oddness with overlap. Bad code on my end. Also, the issue where pulp returns less items than requested appears to be on my end, too.

So... with the latest commit, pagination works as expected with the 3 constraints mentioned earlier. However, I'm only getting `docker_blob` content units. Any ideas?

lcarva:
Filed pulp [issue](https://pulp.plan.io/issues/3931) for the odd behavior mentioned on my previous comment.

We could either not perform the query with an overlap (seems to work, actually), or make an additional request to ensure we're done with fetching results. 

I'll play with the second one for a bit.

twaugh:
The other option is to add a new dockpulp method, perhaps a generator, to just naively page through the results.

Then callers who don't care about having a complete point-in-time view can just step through each result.

Speaking of which, which callers are likely to want naive paging, vs fetch-everything-at-once?

lcarva:
I change the algorithm so if the amount of items returned is less than `limit`, another request is made without overlap. This successfully gets around the pulp issue. The downside is that an additional request is always made. This is probably ok.

Even with pagination though I still can't list contents of `redhat-everything`. It fails after a few pages with 500 error.

lcarva:
500 error was due to sorting, which apparently pulp doesn't like. Removed it and everything seems fine.

PR is ready for a review.

PR #190, opened 2018-08-21
Use absolute timeouts

twaugh:
Signed-off-by: Tim Waugh <twaugh@redhat.com>


twaugh:
Yes, very possibly.

Also note that this PR makes watch_tasks() actually use the provided timeout parameter, which it was not previously doing (it was  ignored).

lcarva:
@twaugh, @breillyr planning on merging this? We should probably have an internal story for tracking testing.

PR #191, opened 2018-09-04
Fix content list pagination with since filter

lcarva:
The previous approach of using '$and' causes Pulp to reply with 500
error indicating $and is an "unknown operator". The fix is to use the
existing filter object for _last_updated field.

### DIFF
--- a/dockpulp/__init__.py
+++ b/dockpulp/__init__.py
@@ -1850,14 +1850,7 @@ class Pulp(object):
                               data=json.dumps(data))
 
         # Adjust filters to avoid picking up content added between page requests
-        existing_last_updated = filter_unit.pop('_last_updated', None)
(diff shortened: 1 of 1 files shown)


dockpulp/__init__.py
```diff
@@ -1850,14 +1850,8 @@ def _collect_repo_units(self, repo_name, filter_unit=None, paginate=True):
                               data=json.dumps(data))
 
         # Adjust filters to avoid picking up content added between page requests
-        existing_last_updated = filter_unit.pop('_last_updated', None)
         lte_filter = {'$lte': time.time()}
```
twaugh: Can remove this line.

lcarva: Fixed, thanks!

PR #194, opened 2018-10-08
Noprotect 5388

breillyr:


tests/test_cli.py
```diff
@@ -41,7 +41,7 @@ def setDebug():
     def isRedirect():
         return
 
-    def createRepo(self, arg1, arg2, desc=None, title=None, protected=None, productline=None,
+    def createRepo(self, arg1, arg2, desc=None, title=None, productline=None, protected=False,
```
twaugh: Why are these switched around?

breillyr:
@twaugh updated, thanks

PR #195, opened 2018-10-22
Schema and List output made default

breillyr:


dockpulp/cli.py
```diff
@@ -614,10 +614,10 @@ def do_list(bopts, bargs, parser):
                       help='show labels')
     parser.add_option('-m', '--manifests', default=False, action='store_true',
                       help='only list manifests and their tags, no blobs')
-    parser.add_option('--schema', default=False, action='store_true',
-                      help='display schema version for each manifest')
-    parser.add_option('--lists', default=False, action='store_true',
-                      help="display manifest lists")
```
twaugh: Does that mean scripts using `--schema` or `--lists` will cause errors?

breillyr: That's a good point. I should probably reroute those options to dummy variables and outputting a deprecation warning...

breillyr: @twaugh made the change, thanks

tests/test_cli.py
```diff
@@ -321,7 +321,7 @@ def test_do_empty(self, mocked_pulp, bargs):
     @patch('dockpulp.Pulp')
     def test_do_list(self, mocked_pulp, caplog, silent):
         bopts = testbOpts()
-        bargs = ['test-repo', '--content', '--details', '--labels', '--lists']
+        bargs = ['test-repo', '--content', '--details', '--labels']
```
twaugh: Could undo this change now, to check the deprecated options still work.

PR #196, opened 2018-12-06
Separate credentials by server

pbabinca:
It's extremely annoying to have to log in each time I switch servers.
Reconciling problems efficiently between environments (especially prod
and stage) results in sensitive passwords being stored in plaintext in
scripts and aliases. This simple patch uses the environment name to
distinguish between credential storage files.


pbabinca:
With #197 in place this will pass flake8 again.

twaugh:
Definitely. Maybe rebase it now #197 is merged?

pbabinca:
Note: this will be backwards incompatible change - anyone who is currently logged in has key/cert in `~/.pulp/pulp/pulp.{cer,key}`.

breillyr:
Should be fine though, as they will just need to login once more. Considering you'd have to do this every time you switched environments before I think that is acceptable.

PR #199, opened 2018-12-20
Quick fix for Pulp manifest list change

breillyr:
Pulp is now providing a dictionary including arch information instead of a simple string digest, dock-pulp needs to accommodate this.


twaugh:
Merging but let's add a unit test when we can.

PR #200, opened 2018-12-21
Remove associated signatures when deleting repo

breillyr:


breillyr:
This is kind of a clumsy solution; ideally I would like to use filters to remove the signature content but I'm concerned about potentially removing signatures that should not be removed. I think I might be able to use regex and remove all signatures that start with `<repoid>@sha256=`, but I am still working on the syntax and there does not seem to be a way for me to test without removing real content.

dockpulp/__init__.py
```diff
@@ -1380,9 +1386,16 @@ def dump(self, pretty=False, paginate=True):
...
+    def emptyRepo(self, repo, sigs=False):
         self.remove_filters(repo)
         log.info('%s emptied' % repo)
+        if sigs:
+            result = self.listRepos(repos=[repo], content=True, paginate=True)[0]
+            repoid = result['docker-id']
+            for manifest in list(result['manifests'].keys()):
+                signature = "%s@%s/signature-1" % (repoid, manifest.replace('@', '='))
+                log.info("Removing signature %s from sigstore", signature)
+                self.remove(SIGSTORE, signature)
```
twaugh: Instead of iterating over each manifest and removing a single signature each time, why not build up a large filter explicitly listing every signature you want to remove, then submit a single remove request?

breillyr: Thanks for the suggestion, that should work perfectly.

dockpulp/__init__.py
```diff
@@ -1354,15 +1357,18 @@ def createRepo(self, repo_id, url, registry_id=None, desc=None, title=None, prot
         self._post('/pulp/api/v2/repositories/', data=json.dumps(stuff))
         return stuff
 
-    def deleteRepo(self, repo, publish=False):
+    def deleteRepo(self, repo, publish=False, sigs=False):
         """Delete a repository; cannot be undone!."""
+        log.info('removing images and manifests from repo %s' % repo)
```
twaugh: Minor: slightly better to use "," instead of "%" for logging functions.

dockpulp/__init__.py
```diff
@@ -1354,15 +1357,18 @@ def createRepo(self, repo_id, url, registry_id=None, desc=None, title=None, prot
...
+        self.emptyRepo(repo, sigs)
         if publish:
-            log.info('removing images and manifests from repo %s' % repo)
-            self.emptyRepo(repo)
             log.info('publishing repo %s twice to remove all content from crane' % repo)
             self.crane(repo, force_refresh=True)
             # Need to publish twice due to order of distributors
             self.crane(repo, force_refresh=True)
+            if sigs:
+                log.info('publishing repo %s to remove related signatures from crane' % SIGSTORE)
```
twaugh: Minor: slightly better to use "," instead of "%" for logging functions.

dockpulp/__init__.py
```diff
@@ -1380,9 +1386,23 @@ def dump(self, pretty=False, paginate=True):
         else:
             return json.dumps(self.listRepos(content=True, paginate=paginate))
 
-    def emptyRepo(self, repo):
+    def emptyRepo(self, repo, sigs=False):
         self.remove_filters(repo)
         log.info('%s emptied' % repo)
+        if sigs:
+            result = self.listRepos(repos=[repo], content=True, paginate=True)[0]
```
twaugh: Why set paginate to True here? It's a single entry isn't it? It feels like either we don't care about pagination, or we should set it to False.

dockpulp/__init__.py
```diff
@@ -1380,9 +1386,23 @@ def dump(self, pretty=False, paginate=True):
...
+            repoid = result['docker-id']
+            signatures = []
+            for manifest in list(result['manifests'].keys()):
```
twaugh: `for manifest in result['manifests']:` is fine here. When iterating over a dictionary the items are the keys, and the the 'keys' view is already iterable so you don't need to convert it to a list.

dockpulp/__init__.py
```diff
@@ -1157,10 +1157,13 @@ def crane(self, repos=[], wait=True, skip=False, force_refresh=False):
...
-            releasekeys = self.release_order.strip().split(",")
             distributors = []
-            for key in releasekeys:
-                distributors.append(self.distributorconf[key])
+            if repo == SIGSTORE:
+                distributors.append(self.distributorconf['iso_distributor_sigstore'])
+            else:
+                releasekeys = self.release_order.strip().split(",")
+                for key in releasekeys:
+                    distributors.append(self.distributorconf[key])
```
twaugh: Would be nice to have test coverage here (actually for any of crane()).

dockpulp/__init__.py
```diff
@@ -1975,6 +1995,21 @@ def remove(self, repo, img):
...
+                    'filters': {
+                        'unit': {
+                            "$or": [{'name': img}]
+                        }
+                    }
+                },
+                'limit': 1,
+                'override_config': {}
+            }
+
```
twaugh: Would be nice to have test coverage here (actually for any of remove()).

dockpulp/__init__.py
```diff
@@ -1997,13 +2032,15 @@ def remove(self, repo, img):
...
-    def remove_filters(self, repo, filters={}, v1=True, v2=True):
+    def remove_filters(self, repo, filters={}, v1=True, v2=True, sigs=False):
         """Remove content from a repo according to filters."""
         type_ids = []
         if v1:
             type_ids.append(V1_C_TYPE)
         if v2:
             type_ids.extend([V2_C_TYPE, V2_BLOB, V2_TAG, V2_LIST])
+        if sigs:
+            type_ids.append(SIG_TYPE)
```
twaugh: Would be nice to have test coverage for this.

breillyr:
@twaugh thanks, for the input, updated the PR.

tests/test_pulp_instance.py
```diff
@@ -55,6 +55,14 @@ def pulp(tmpdir):
                 "docker_rsync":{
                     "distributor_type_id": "docker_rsync_distributor",
                     "distributor_config": {}
+                },
+                "iso_distributor_sigstore": {
+                    "distributor_id": "iso_distributor",
+                    "distributor_type_id": "iso_distributor",
+                    "distributor_config": {
+                        "rsync_extra_args": ["--exclude", "PULP_MANIFEST"]
```
twaugh: I think this still correct even in light of last week's outage(?)

tests/test_pulp_instance.py
```diff
@@ -755,6 +763,24 @@ def test_createRepo(self, pulp, repo_id, url, registry_id, distributors, product
             else:
                 assert response['notes']['include_in_download_service'] == "False"
 
+    @pytest.mark.parametrize('repo_id', ['redhat-sigstore'])
```
twaugh: Is it hard to also add a test for some other repo?

PR #205, opened 2019-01-21
Better logging for Pulp failures

breillyr:
If dockpulp fails to parse the json it should not simply discard the response.


twaugh:
I feel like we're doing something backwards here. We should be checking the status code before looking at the content in the response. If we get a 403 response, for example, nothing assures us we'll get a JSON document in the response, but we might. So we might raise DockPulpLoginError, but we might raise DockPulpError instead, depending on exactly what the content looks like.

I think we should only be trying to decode the content `if answer.ok`. If `answer.ok` and decoding the content fails, or if `!answer.ok`, let's log the status code and raw content.

Then if `!answer.ok` we should raise the appropriate exception.

breillyr:
@twaugh thanks for the review, I've updated the PR.

dockpulp/__init__.py
```diff
@@ -158,17 +158,27 @@ def __call__(self, meth, api, **kwargs):
...
+                r = answer.json()
+                log.debug('raw response data:')
+                log.debug(pprint.pformat(r))
+            except ValueError:
+                log.debug('status code:')
+                log.debug(answer.status_code)
+                log.debug('raw response data:')
+                log.debug(pprint.pformat(answer.content))
+                raise errors.DockPulpError('No valid JSON in Pulp response')
+        elif not answer.ok:
```
twaugh: Just `else:` I guess. Another way to write it would be to have the `if not answer.ok` test first, where all paths lead to raising an exception, and then the "happy path" can continue unindented:
```
if not answer.ok:
    # log status code and content...
    # check which exception to raise
    if answer.status_code == requests.codes.forbidden:
        raise errors.DockPulpLoginError(...)
    self._error(...)  # raises exception

# successful response, so decode content...
try:
    r = answer.json()
    # etc
```

dockpulp/__init__.py
```diff
@@ -158,17 +158,27 @@ def __call__(self, meth, api, **kwargs):
...
-            raise errors.DockPulpError('No content in Pulp response')
-        if answer.status_code == 403:
-            raise errors.DockPulpLoginError('Received 403: Forbidden')
-        elif answer.status_code >= 400:
-            self._error(answer.status_code, url)
-        elif answer.status_code == 202:
+            log.debug(answer.content)
+            if answer.status_code == 403:
+                raise errors.DockPulpLoginError('Received 403: Forbidden')
+            elif answer.status_code >= 400:
```
twaugh: This test looks redundant because it's testing the same thing `answer.ok` is.

breillyr:
@twaugh thanks, updated.

PR #206, opened 2019-01-24
Signature removal now removes all incremented signatures

breillyr:


dockpulp/__init__.py
```diff
@@ -1389,7 +1389,9 @@ def deleteSignatures(self, repo):
         repoid = result['docker-id']
         signatures = []
         for manifest in result['manifests']:
-            signature = {'name': "%s@%s/signature-1" % (repoid, manifest.replace(':', '='))}
+            signature = {'name':
+                         # signatures can increment, we always want to remove all of them
+                         {'$regex': "%s@%s/signature-.*" % (repoid, manifest.replace(':', '='))}}
```
twaugh: Looks safe since neither `repoid` or the manifest name are allowed characters which are special for regex, so no `re.escape()` needed.

dockpulp/__init__.py
```diff
@@ -2023,7 +2025,7 @@ def remove(self, repo, img, sigs=False):
                     'type_ids': [SIG_TYPE],
                     'filters': {
                         'unit': {
-                            "$or": [{'name': img}]
+                            'name': img
```
twaugh: Looks like this is just a small improvement unrelated to using regexes. Looks good.

PR #207, opened 2019-02-12
 sigstore now publishes all associated repos 

breillyr:


dockpulp/__init__.py
```diff
@@ -1166,12 +1166,13 @@ def crane(self, repos=[], wait=True, skip=False, force_refresh=False):
...
         for repo in repos:
             distributors = []
+            releasekeys = self.release_order.strip().split(",")
             if repo == SIGSTORE:
                 distributors.append(self.distributorconf['iso_distributor_sigstore'])
-            else:
-                releasekeys = self.release_order.strip().split(",")
-                for key in releasekeys:
-                    distributors.append(self.distributorconf[key])
+                # need to cut off last distributor in release order for sigstore
```
rcerven: why?

breillyr:
Updated the pull request for a better way of managing this. Needs a rh-dockpulp change, but 'sig_release_order' is defined in the same way that 'release_order' is, just specific to the dists that are published for sigstore specifically.

PR #209, opened 2020-04-15
Remove image signatures in batches

midnightercz:
To avoid mongo limitation which results in error: BSON document too large
we need to limit number of image signatures removed in one request from
a repository

### DIFF
--- a/dockpulp/__init__.py
+++ b/dockpulp/__init__.py
@@ -17,6 +17,7 @@ import six
 import atexit
 from datetime import datetime
 from six.moves import configparser
+from six.moves import zip_longest
 import hashlib
 import logging
 import os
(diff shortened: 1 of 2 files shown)


dockpulp/__init__.py
```diff
@@ -1400,14 +1409,19 @@ def deleteSignatures(self, repo):
...
+            filtered_batch = [b for b in batch if b]
+            filters = {
+                'unit': {
+                    "$or": filtered_batch
+                }
             }
-        }
-        log.debug("Removing signatures from sigstore: %s", signatures)
-        self.remove_filters(SIGSTORE, filters, v1=False, v2=False, sigs=True)
+            log.debug("Removing signatures from sigstore [%d/%d]: %s",
```
querti, line 1421: Wouldn't the first log say [0/5], and the last [4/5]?

midnightercz: hmm, good point

PR #211, opened 2020-06-05
Not wait for tasks in terminal states

emilyzheng:
Task watch function returns immediately when task is in state
skipped or canceled.

### DIFF
--- a/dockpulp/__init__.py
+++ b/dockpulp/__init__.py
@@ -2404,7 +2404,7 @@ class Pulp(object):
                 log.info('subtask skipped')
                 return t
             elif t['state'] == 'canceled':
-                raise errors.DockPulpError('subtask canceled, please retry')
+                raise errors.DockPulpError('subtask canceled')
             else:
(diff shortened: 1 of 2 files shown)


dockpulp/__init__.py
```diff
@@ -2400,6 +2400,11 @@ def watch(self, tid, timeout=None, poll=5):
                 log.debug('traceback from subtask:')
                 log.debug(t['traceback'])
                 raise errors.DockPulpTaskError(t['error'])
+            elif t['state'] == 'skipped':
+                log.info('subtask skipped')
+                return t
+            elif t['state'] == 'canceled':
+                raise errors.DockPulpError('subtask canceled, please retry')
```
rohanpm, line 2407: Why do we suggest retrying here, while we don't suggest retrying in the error or timeout cases?

I would prefer if it just stated the facts and didn't suggest the user retry, it doesn't seem like this code can possibly know whether that's the right thing to do.

PR #212, opened 2020-08-28
Report error when --server isn't specified

emilyzheng:
'qa' was the default value of server. It could be confusing if it's not noticed. E.g. a user may login to the default environment without specifying --server, but not aware it's 'qa', then get error of 'need to log in' when doing other operations with a --server other than 'qa'. Make --server required to avoid the confusion.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -485,3 +485,9 @@ class TestCLI(object):
                 cli.do_update(bopts, bargs)
         else:
             assert cli.do_update(bopts, bargs) is None
+
+    def test_no_server(self, capsys):
+        with pytest.raises(SystemExit):
+            cli.main(args=['list'])
+        out, err = capsys.readouterr()
+        assert "Please specify --server" in err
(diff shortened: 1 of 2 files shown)


lipoja (approved):

tests/test_cli.py
```diff
@@ -485,3 +485,9 @@ def test_do_update(self, mocked_pulp, bargs):
                 cli.do_update(bopts, bargs)
         else:
             assert cli.do_update(bopts, bargs) is None
+
+    def test_no_server(self, capsys):
+        with pytest.raises(SystemExit):
+            cli.main(args=['list'])
+        out, err = capsys.readouterr()
+        assert "Please specify --server" in err
```
lipoja, line 493: This works as expected. I just wanted to find something and the only thing that I found is missing "." at the end of "Please specify --server" ;-)

midnightercz (approved):

querti (approved):

PR #217, opened 2022-12-12
Update RPM spec file to work with RHEL-8 / python 3 builds

MichalHaluza:
This also simplifies how the python package is built and how the files are collected in order to create the RPM package. On RHEL-6 and RHEL-7, the RPMs will end up looking mostly the same (they will include additional egg-info files). On RHEL-8 as well as modern Fedoras the package will become buildable and will work with python 3.

### DIFF
--- a/dockpulp.spec
+++ b/dockpulp.spec
@@ -1,21 +1,34 @@
+%if 0%{?rhel} && 0%{?rhel} <= 7
+%define python python
+%else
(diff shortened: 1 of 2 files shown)


yashvardhannanavati:
lgtm as far as I can understand :)

emilyzheng (approved):

chandwanitulsi (approved):

PR #220, opened 2023-03-13
Update spec file and remove RHEL6

amcmahon-rh:
It it turns out that the usual installation method  `setup.py install --record=INSTALLED_FILES` records just the files, but not the directories which contain those files. This results in empty egg-info directories left in /usr/lib/python3.6/site-packages after an RPM with python module is updated to a newer version.

This change ensures egg-info directories are included and also removes the RHEL6 package target as we are moving away from it.

### DIFF
--- a/dockpulp.spec
+++ b/dockpulp.spec
@@ -1,9 +1,3 @@
-%if 0%{?rhel} && 0%{?rhel} <= 7
-%define python python
-%else
-%define python %{__python3}
-%endif
-
 Name:		dockpulp
 Version:	1.72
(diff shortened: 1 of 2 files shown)


amcmahon-rh:
> @amcmahon-rh: setup.py also defines 'conf/dockpulp.conf', 'conf/dockpulpdistributors.json' . These need to be packaged as well.

Looking at the packages, I can't see either of these config files being included. It appears the way `package_data` works relative to the package directory, rather than when setup.py is located.

MichalHaluza (approved):

PR #221, opened 2023-03-21
Drop the py26 and py27 in tox.ini file [RHELDST-15248]

LinglingXiang:
1.Drop the travil and add the github action
2.Drop the py26 and py27


LinglingXiang:
@rbikar , please help review again,  thanks

rbikar (approved):

